PR #2, opened 2020-12-12
Added some settings to local config

effernity:
Added to local config (mega and MKS):
RMS current RA and DEC
Speed of RA and DEC
Acceleration of RA and DEC


effernity:
Errors found

PR #8, opened 2020-12-13
PH firmware/develop branch integration

philharrisathome:
Brought all previous PR into line with OpenAstroTracker-Firmware repository.

1. Rationalized & documented interrupt (Atmega) and task (ESP32) management, fixed kernel panic and enabling gyro & GPS on ESP32.
2. Added support for SSD1306 display,
3. Fixed minor compiler warnings on ESP32.
4. Reduced bandwidth of gyro for smoother tilt/pitch calibration. Added documentation and named constants.
5. Added configuration parameters for serial baudrate according to external controller (Stellarium, ASCOM, etc.)
6. Reinstated brightness control
7. Use named storage locations in EPROM

Tested on ESP32. Compiled (but not tested) on Atmega.

### DIFF
--- a/src/LcdMenu.cpp
+++ b/src/LcdMenu.cpp
@@ -9,24 +9,27 @@
 // You add a string and an id item and this class handles the display and navigation
 // Create a new menu, using the given number of LCD display columns and rows
 #if DISPLAY_TYPE == DISPLAY_TYPE_LCD_KEYPAD
-LcdMenu::LcdMenu(byte cols, byte rows, int maxItems) : _lcd(LCD_PIN8, LCD_PIN9, LCD_PIN4, LCD_PIN5, LCD_PIN6, LCD_PIN7)
+LcdMenu::LcdMenu(byte cols, byte rows, int maxItems) : _lcd(LCD_PIN8, LCD_PIN9, LCD_PIN4, LCD_PIN5, LCD_PIN6, LCD_PIN7),
+  _cols(cols), _rows(rows), _maxItems(maxItems),
+  _charHeightRows(1)  // 1 character = 1 row
 {
-  _cols = cols;
-  _rows = rows;
-  _maxItems = maxItems;
 }
 #elif DISPLAY_TYPE == DISPLAY_TYPE_LCD_KEYPAD_I2C_MCP23008 || DISPLAY_TYPE == DISPLAY_TYPE_LCD_KEYPAD_I2C_MCP23017
-LcdMenu::LcdMenu(byte cols, byte rows, int maxItems) : _lcd(0x20)
+LcdMenu::LcdMenu(byte cols, byte rows, int maxItems) : _lcd(0x20),
+  _cols(cols), _rows(rows), _maxItems(maxItems),
+  _charHeightRows(1)  // 1 character = 1 row
 {
   #if DISPLAY_TYPE == DISPLAY_TYPE_LCD_KEYPAD_I2C_MCP23017
   _lcd.setMCPType(LTI_TYPE_MCP23017);
   #elif DISPLAY_TYPE == DISPLAY_TYPE_LCD_KEYPAD_I2C_MCP23008
   _lcd.setMCPType(LTI_TYPE_MCP23008);
   #endif
-
-  _cols = cols;
-  _rows = rows;
-  _maxItems = maxItems;
+}
+#elif DISPLAY_TYPE == DISPLAY_TYPE_LCD_JOY_I2C_SSD1306
+LcdMenu::LcdMenu(byte cols, byte rows, int maxItems) :
+  _cols(cols), _rows(rows), _maxItems(maxItems),
+  _charHeightRows(2)  // 1 character = 2 rows (2x8 pixels)
+{
 }
 #endif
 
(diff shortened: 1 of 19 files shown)


src/LcdMenu.cpp
```diff
@@ -122,14 +134,18 @@ void LcdMenu::setBacklightBrightness(int level, bool persist)
 {
   _brightness = level;
 
-  //LOGV2(DEBUG_INFO, F("LCD: Writing %d as brightness"), _brightness  );
-  // analogWrite(10, _brightness);
-  //LOGV2(DEBUG_INFO, F("LCD: Wrote %d as brightness"), _brightness  );
+  #if DISPLAY_TYPE == DISPLAY_TYPE_LCD_KEYPAD && defined(LCD_BRIGHTNESS_PIN)
+    analogWrite(LCD_BRIGHTNESS_PIN, _brightness);
```
ClutchplateDude: Brightness does not work on a Mega with a standard LCD Shield, we should disable this by default and make the CAL menu entry disable-able through defines.  As soon as you change from 255 down it goes dark, then once you get to around 194 it crashes hard.

src/b_setup.hpp
```diff
@@ -47,52 +49,52 @@ Mount mount(RA_STEPS_PER_DEGREE, DEC_STEPS_PER_DEGREE, &lcdMenu);
...
-/////////////////////////////////
+// This is the task for simulating periodic interrupts on ESP32 platforms. 
+// It should do very minimal work, only calling Mount::interruptLoop() to step the stepper motors as needed.
+// This task function is run on Core 0 of the ESP32 and never returns
 void IRAM_ATTR stepperControlTask(void* payload)
 {
   Mount* mount = reinterpret_cast<Mount*>(payload);
   for (;;) {
     mount->interruptLoop();
+    vTaskDelay(1);  // 1 ms 	// This will limit max stepping rate to 1 kHz
```
ClutchplateDude, line 84: Is there a specific reason you added this back in? We did some experimentation and found that removing this made the stepper run better....

ClutchplateDude (commented):
Great PR, just some more clarifications needed....

src/b_setup.hpp
andre-stefanov: Just to add some context. we introduced this delay once because the esp32 watchdog caused a crash thinking that the core is never idling (i mean ... its true but its not an issue). But since we are using `disableCore0WDT()`, we don't have to delay here. The only thing to be still verified is if this somehow influences WiFi in a bad way since it is running on Core 0 per default iirc.

src/LcdMenu.cpp
andre-stefanov: We really have to investigate the crash at some point ... it should actually work.

ClutchplateDude: I have already spent too much time on this, back when I first try to add it... :-)

andre-stefanov: Unfortunatelly analogWrite does not compile for me for the esp32. how did you build it? did you add some library which delegates analogWrite to LEDC?

andre-stefanov (changes requested):

src/LcdMenu.cpp
ClutchplateDude: Well, DISPLAY_TYPE_LCD_KEYPAD also is not available with ESP32, so it probably wouldn't work anyway? analogWrite is a very basic Arduino function, it should exist on ESP32

andre-stefanov: thats exactly what i thought as well ... but unfortunatelly analogWrite requires specific hardware capabilities which are apparently not supported by esp32: https://github.com/espressif/arduino-esp32/issues/4

i think if we leave `DISPLAY_TYPE_LCD_KEYPAD` unsupported for esp32, we should handle this with an `#error` message in configs. because this new code in LcdMenu.cpp is actually meant for esp + lcd keypad but it will not compile.

andre-stefanov (changes requested):
We have to fix the esp32 lcd keypad compiler error before merging (probably just am appropriate error message would be enough).

src/LcdMenu.cpp
philharrisathome: Unfortunately I don't an ATmega to test this. I suggest to comment it out for now and add a TODO mentioning the crash.

src/b_setup.hpp
philharrisathome: I've added this back to allow some time on Core 0 for other tasks, like Wifi. stepperControlTask() is running at a slightly elevated priority, without this it will consume all time available on Core 0. In the FreeRTOS build for ESP32 1 ms is the scheduler period, so cannot go lower. I didn't see any issues with stepper smoothness - I will check again.

src/LcdMenu.cpp
philharrisathome: Apologies, I missed this. I have an additional PR coming that includes configuration checking that disallows the ESP32/ DISPLAY_TYPE_LCD_KEYPAD configuration. An aside, but in the past I have used https://github.com/ERROPiX/ESP32_AnalogWrite to implement analogWrite on ESP32.

andre-stefanov (approved):

src/b_setup.hpp
andre-stefanov: i will have to investigate if this is still an issue ...

PR #9, opened 2020-12-18
Fixing steps per degree

philharrisathome:
Reviewed and documented microsteps management in Mount and related classes. 

- Some rationalization of driver configuration & calculation code in Mount.
- Identified issue with switching of microstep configuration between slew/track/guide that results in incorrect position() readings. Forced single microstep values for all three drivers to ensure correct behavior, but at possible loss of fine guiding accuracy.
- Intention is to resolve this through refactoring into HAL such that an acccurate position () value can be maintained despite dynamic microstep switching.
- Added validation of all known system configurations and required pin assignments into Configuration_adv.hpp 
- Possible breaking change for local configurations: Renamed DRIVER_TYPE_GENERIC to DRIVER_TYPE_A4988_GENERIC


philharrisathome:
My github workflow needs some work :disappointed:. I have added a new PR that correctly merges the "fixing-steps-per-degree" and "new-epromstore" into develop without conflicts. You may want to decline this one.

PR #11, opened 2020-12-19
'fixing-steps-per-degree' and 'new-epromstore' merged into 'develop'

philharrisathome:
This PR combines the previous 'fixing-steps-per-degree' and 'new-epromstore' PRs, with the branches correctly merged into upstream/develop. It may be easier to decline the two previous PRs and use this one instead for easier merging.

New EEPROMStore implementation, abstracting all scaling and low-level store management away from Mount etc.

- EEPROMStore is now guaranteed to return a valid value for any item, regardless of whether it is present in NVM or not. This simplifies Mount considerably.
- Implementation is transaction-based such that flag setting & item value storing occurs in an atomic fashion through commit() (at least on ESP32).
- Quietened noisy CTORs in DayTime and added assignment operator.

Reviewed and documented microsteps management in Mount and related classes.

Some rationalization of driver configuration & calculation code in Mount.

- Identified issue with switching of microstep configuration between slew/track/guide that results in incorrect position() readings. Forced single microstep values for all three drivers to ensure correct behavior, but at possible loss of fine guiding accuracy.
- Intention is to resolve this through refactoring into HAL such that an acccurate position () value can be maintained despite dynamic microstep switching.
- Added validation of all known system configurations and required pin assignments into Configuration_adv.hpp
- Possible breaking change for local configurations: Renamed DRIVER_TYPE_GENERIC to DRIVER_TYPE_A4988_GENERIC


Configuration_adv.hpp
```diff
@@ -348,12 +408,188 @@
...
+  #if !defined(DEC_STEP_PIN) || !defined(DEC_DIR_PIN) || !defined(DEC_EN_PIN) || !defined(DEC_DIAG_PIN) || !defined(DEC_SERIAL_PORT_TX) || !defined(DEC_SERIAL_PORT_RX)
+     // Required pin assignments missing
+  #endif
 #endif
-#if RA_RMSCURRENT > 2000 || DEC_RMSCURRENT > 2000
-#error "Do you really want to set the RMS motorcurrent above 2 Ampere? Thats almost 3A peak! Delete this error if you know what youre doing" 
+
+#if (RA_DRIVER_TYPE == DRIVER_TYPE_ULN2003)
+  #if !defined(RA_IN1_PIN) || !defined(RA_IN2_PIN) || !defined(RA_IN3_PIN) || !defined(RA_IN4_PIN)
+     // Required pin assignments missing
```
ClutchplateDude, line 543: Don't we want to generate a #error here?

ClutchplateDude (approved):

src/DayTime.cpp
```diff
@@ -69,20 +69,31 @@ DayTime::DayTime(const DayTime &other)
 
 DayTime::DayTime(int h, int m, int s)
 {
-  LOGV4(DEBUG_INFO, "DayTime ctor(%d, %d, %d)", h, m, s);
+  //LOGV4(DEBUG_INFO, "DayTime ctor(%d, %d, %d)", h, m, s);
```
andre-stefanov: We should not leave commented code. either we need the logs here or we should remove them completely. Additionally the logs are managed by log level anyway to prevent to much verbosity.

src/DayTime.cpp
```diff
@@ -69,20 +69,31 @@ DayTime::DayTime(const DayTime &other)
...
-  LOGV2(DEBUG_INFO, "DayTime ctor(%f)", timeInHours);
+  //LOGV2(DEBUG_INFO, "DayTime ctor(%f)", timeInHours);
   long sgn = fsign(timeInHours);
   timeInHours = abs(timeInHours);
   totalSeconds = sgn * round(timeInHours * 60 * 60);
-  LOGV4(DEBUG_INFO, "DayTime ctor(%f) -> %l -> %s", timeInHours, totalSeconds, ToString());
+  //LOGV4(DEBUG_INFO, "DayTime ctor(%f) -> %l -> %s", timeInHours, totalSeconds, ToString());
+}
+
+DayTime& DayTime::operator=(DayTime const& other) // copy assignment
```
andre-stefanov: the assignment operator logic does not match the copy constructor (hourWrap is missing in copy constructor)

src/EPROMStore.cpp
```diff
@@ -4,152 +4,659 @@
...
 }
 
 #else
 
 // Initialize the EEPROM storage in a platform-independent abstraction
-void EPROMStore::initialize()
+void EEPROMStore::initialize()
 {
-  LOGV1(DEBUG_EEPROM, F("EEPROM[Uno/Mega]: Startup "));
+  LOGV1(DEBUG_EEPROM, F("EEPROM[Uno/Mega]: Startup"));
```
andre-stefanov: We don't support UNO anymore. We should simply remove the board specification here because this is the "else" case and could theoretically be used even on other boards as mega. The used board should be clear at runtime anyway and there is no real need to check if #ifs are working properly :)

src/EPROMStore.cpp
```diff
@@ -4,152 +4,659 @@
...
+
+  // Read the magic marker byte and state
+  uint16_t marker = readUint16(MAGIC_MARKER_AND_FLAGS_ADDR);
+  LOGV2(DEBUG_EEPROM, F("EEPROM[Uno/Mega]: Magic Marker: %04X"), marker);
+
+  LOGV1(DEBUG_INFO|DEBUG_EEPROM, ((marker & MAGIC_MARKER_MASK) == MAGIC_MARKER_VALUE) ? F("EEPROM has values") : F("EEPROM does NOT have values"));
```
andre-stefanov: i would make these logging commands multiline because it's not easy to read with that amount of comparators and ternary operator

src/EPROMStore.cpp
```diff
@@ -4,152 +4,659 @@
...
+uint8_t EEPROMStore::read(uint8_t location)
 {
   uint8_t value;
   value = dummyEepromStorage[location];
-  LOGV3(DEBUG_EEPROM, F("EEPROM[DUMMY]: Read %x from %d"), value, location);
+  LOGV3(DEBUG_EEPROM, F("EEPROM[DUMMY]: Read %02X from %d"), value, location);
   return value;
 }
 
 #elif defined(ESPBOARD)
```
andre-stefanov, line 45: could we split the different platform EEPROM implementations in two separate cpp files? this would improve readability and also make the HAL migration for EEPROM easier later.

Configuration_adv.hpp
philharrisathome: Yes. I have added errors to all these. Somehow it missed the commit but it is incoming.

philharrisathome: Done. I have moved the #include of Configuration_pins into Configuration_adv so they are always picked up by the validation code. I think it would make sense to combine Configuration_pins and Configuration_local_xxx as the pin assignments are often specific to a physical hardware build (I've had to recreate Configuration_pins a couple of times after it got overwritten by upstream changes). Then the configuration construction order would be Configuration_local_xxx > Configuration > Configuration_adv, each able override the next. Let me know if you're happy for me to do this.

philharrisathome: Also, I saw on reddit that people were having trouble with storing of roll offsets. I have confirmed that this works correctly (at least in this PR).

src/Mount.cpp
```diff
@@ -107,19 +67,17 @@ const float siderealDegreesInHour = 14.95904348958;
...
-  _stepsPerRADegree = stepsPerRADegree * SET_MICROSTEPPING; // hier
-  #else
-  _stepsPerRADegree = stepsPerRADegree;
-  #endif
-  #if DEC_DRIVER_TYPE != DRIVER_TYPE_ULN2003
-  _stepsPerDECDegree = stepsPerDECDegree * DEC_SLEW_MICROSTEPPING;
-  #else
-  _stepsPerDECDegree = stepsPerDECDegree;
+Mount::Mount(LcdMenu* lcdMenu) :
+  _stepsPerRADegree(RA_STEPS_PER_DEGREE),   // u-steps per degree when slewing
```
andre-stefanov, line 68: its not always microsteps ... if the driver is set to full step mode it would be just steps. also the ULN2003 does not support microstepping

src/Mount.cpp
```diff
@@ -107,19 +67,17 @@ const float siderealDegreesInHour = 14.95904348958;
...
+  _stepsPerDECDegree(DEC_STEPS_PER_DEGREE)  // u-steps per degree when slewing
+  #if AZIMUTH_ALTITUDE_MOTORS == 1
+    , _stepsPerAZDegree(AZIMUTH_STEPS_PER_REV / 360),
```
andre-stefanov, line 71: we are using 2 different ways to calculate steps per deg. for RA and DEC it gets calculated completelly in adv configs, for alt and az it gets divided by 360 here ... consistency would make it cleaner

src/Mount.cpp
```diff
@@ -107,19 +67,17 @@ const float siderealDegreesInHour = 14.95904348958;
...
+    _stepsPerALTDegree(ALTITUDE_STEPS_PER_REV / 360),
+    _azAltWasRunning(false)
   #endif
+{
+   _instance = this;
```
andre-stefanov: if i see it correctly, we don't use this singleton anywhere ... could be deleted

src/Mount.cpp
```diff
@@ -473,12 +179,12 @@ void Mount::writePersistentData(int which, long val)
...
 #if RA_STEPPER_TYPE == STEPPER_TYPE_28BYJ48
-void Mount::configureRAStepper(byte stepMode, byte pin1, byte pin2, byte pin3, byte pin4, int maxSpeed, int maxAcceleration)
+void Mount::configureRAStepper(byte pin1, byte pin2, byte pin3, byte pin4, int maxSpeed, int maxAcceleration)
 {
 #if NORTHERN_HEMISPHERE
-  _stepperRA = new AccelStepper(stepMode, pin4, pin3, pin2, pin1);
+  _stepperRA = new AccelStepper((RA_SLEW_MICROSTEPPING == 1) ? AccelStepper::FULL4WIRE : AccelStepper::HALF4WIRE, pin4, pin3, pin2, pin1);
 #else
-  _stepperRA = new AccelStepper(stepMode, pin1, pin2, pin3, pin4);
+  _stepperRA = new AccelStepper((RA_SLEW_MICROSTEPPING == 1) ? AccelStepper::FULL4WIRE : AccelStepper::HALF4WIRE, pin1, pin2, pin3, pin4);
```
andre-stefanov, line 182: mh ... luckily all this will change with HAL anyway :)

src/Mount.cpp
```diff
@@ -1281,13 +950,19 @@ void Mount::guidePulse(byte direction, int duration) {
 // The first time with EAST, second with WEST and then with 0.
 /////////////////////////////////
 void Mount::runDriftAlignmentPhase(int direction, int durationSecs) {
-  // Calculate the speed at which it takes the given duration to cover 400 steps.
-  float speed = 400.0 / durationSecs;
+
+  float const numArcMinutes(5.3);
```
andre-stefanov, line 950: what is this?

src/Mount.cpp
```diff
@@ -2299,7 +1991,7 @@ float Mount::getSpeed(int direction) {
 //
 // This code tells the steppers to what location to move to, given the select right ascension and declination
 /////////////////////////////////
-void Mount::calculateRAandDECSteppers(float& targetRA, float& targetDEC) {
+void Mount::calculateRAandDECSteppers(DayTime const& ra, Declination const& dec, long& targetRASteps, long& targetDECSteps) const {
```
andre-stefanov, line 1990: with HAL refactoring we should stop using DayTime as type for RA position. We should create a DEC and RA model classes.

andre-stefanov (changes requested):

src/DayTime.cpp
philharrisathome: Agreed. This one is really at the wrong level of severity so heavily pollutes the logs. I wasn't sure if anyone was using it for ongoing DayTime development, so left it in for now.

src/DayTime.cpp
philharrisathome: Agreed. I will review the existing implementation and fix this.

src/EPROMStore.cpp
philharrisathome: Fair enough. Will update.

src/EPROMStore.cpp
philharrisathome: Fair enough. Will update.

src/EPROMStore.cpp
philharrisathome: We could do. It's not too bad right now as the differences are very small and clear. Its a trade off between proliferation of files and clarity of implementation. The code in EEPROM is so boring its probably not worth it right now - the axis controllers are perhaps a different story.

src/Mount.cpp
philharrisathome: Well, I just labelled it microstepping to differentiate from full steps. Would be more correct to state "includes microstepping when microstepping enabled". I just consider half steps as large microsteps and full-steps as the degenerate unity case :-). This is also consistent with the datasheets for the non-ULN2003 drivers. If we use engineering units for the HAL drivers then a lot of this will become clearer.

src/Mount.cpp
philharrisathome: Agreed. I haven't reviewed the AZ/ALT implementation yet.

src/Mount.cpp
philharrisathome: Agreed. I think this is legacy of very old interrupt handling on ATmega.

src/Mount.cpp
philharrisathome: Yeah. I moved it in from setup to make it clearer - one step closer to the HAL. I am trying to do these changes incrementally - hopefully they will accumulate into natural groupings that can be easily copied to the HAL implementation, otherwise we will be changing everything everywhere and merging will be a nightmare.

src/Mount.cpp
philharrisathome: 400 steps was hardcoded but didn't take into account the current microstepping mode. The 5.3 arc-minutes is from the function documentation which states 400 steps is 5.3 arc-minutes. So now we get 5.3 arc-minutes of motion regardless of microstepping mode. I don't know how 5.3 arc-minutes was chosen :-(.

src/Mount.cpp
philharrisathome: Agreed. There is a Declination class, dunno why there isn't an RA equivalent.

src/EPROMStore.cpp
andre-stefanov: mh fair enough :)

src/Mount.cpp
andre-stefanov: yeah probably

src/Mount.cpp
andre-stefanov: weird ... don't know either what we should do about this one at the moment (if at all)

ClutchplateDude: It's just an arbitrary distance that can be used with a "normal" lens (like 100-200mm) that causes star trails about 1/5 of the camera view.

src/DayTime.cpp
ClutchplateDude: Yeah, that one should have been VERBOSE.. oh well, it's debugged and working, so removal is fine :-)

src/DayTime.cpp
ClutchplateDude: Actually, we can remove hourWrap from this class, since I derived the Declination class (needs a different wrap value). The wrapping is now handled in the virtual checkHours() method.

src/EPROMStore.cpp
ClutchplateDude: You cannot change %x to %02X. The LOGVx family uses our own formatter and it does not support the full printf() format spec. I think I had to implement this because one of the platforms didn't support sprintf(). If you can get sprintf() to work on our platforms, we could replace formatArg() in Utility.cpp with a sprintf() call.

src/Mount.cpp
ClutchplateDude: RA would have no extra functionality to DayTime. But we should probably just derive it for clarity.

src/Mount.cpp
ClutchplateDude: I think the Meade processor used to pull this... but we changed that and injected it instead. So, yeah, remove.

src/EPROMStore.cpp
philharrisathome: Done.

src/EPROMStore.cpp
philharrisathome: Done

src/EPROMStore.cpp
philharrisathome: Done. So much for the last minute cosmetic change :-(

src/Mount.cpp
philharrisathome: Done.

src/DayTime.cpp
philharrisathome: I've added hourWrap to the copy constructor - good catch! The other constructors initialize from declaration.
@ClutchplateDude Can you please take a look at hourWrap? DayTime::checkHours() is still using it.

src/DayTime.cpp
philharrisathome: Done - removed.

ClutchplateDude:
As mentioned hourWrap is not needed to override. Can just hardcode to 24. Checkhours is virtual so...

ClutchplateDude (approved):

src/DayTime.cpp
ClutchplateDude: Just to verify, the hourWrap was eliminated, right?

ClutchplateDude: nvm, I saw the commit...

src/Mount.cpp
ClutchplateDude: This can be done in another PR, probably with Camerons input...

Configuration_adv.hpp
andre-stefanov: Mh ... i somehow miss the commit you mentioned with added `#error`

PR #13, opened 2020-12-23
Fix guide pulse handling

pmneo:


ClutchplateDude:
I have ported this to the repo, but since it also requires an ASCOM Driver update so as not to break Windows I need to also create a ASCOM Driver change and commit both of these.
Thanks for the contribution.

PR #14, opened 2020-12-23
Automatic HA calculation from Indi Driver Infos

pmneo:


CTetford:
Overall this looks good to go with a couple minor changes.

This is a good basis for re-working the GPS code so it isn't dependent on the LCD menu files.

Ideally if the system has GPS functionality it should use that, then fall back on this if it's connected to INDI and doesn't have a good satellite lock.

ClutchplateDude (approved):

PR #15, opened 2020-12-23
Added support for TMC2209 connection testing

pmneo:


andre-stefanov:
was implemented in #32

PR #17, opened 2020-12-23
Run DEC and RA Pulse with independent duration

pmneo:
Otherwise a RA pulse or DEC pulse will cut or extend the previous requested pulse


pmneo:
It also may be better to use a own accel stepper instance, because of different microstepping? In this case the stepper position may result in a wrong information.

ClutchplateDude:
I implemented a variation of this directly. Please use coding standards from the file you edit (no spaces inside parens, for example).

PR #18, opened 2020-12-30
Revised configuration files

philharrisathome:
A fairly expansive change set to rationalize management of include files. Now there is only one configuration file that requires customization by the user (Configuration_local_*.hpp). Configuration_pins.hpp has been merged into Configuration_local_*.hpp and subsequently deleted. For parameters that are not explicitly given, safe defaults are provided by Configuration.hpp. Required missing parameters result in descriptive #error messages. Parameter checking (previous PR) that was in Configuration_adv.hpp is now in Configuration.hpp and expanded to cover more configuration items (e.g. Wifi credentials). 

A side effect is that it may break users' existing local configuration customizations, but longer term it should bring configuration file stability (I know I've overwritten my hardware configuration several times :-( ).

Configuration accumulates in order across 3 files: Configuration_local_*.hpp, Configuration.hpp, and Configuration_adv.hpp.
 - Configuration_local_*.hpp (see below) captures the local (primarily hardware) configuration.
 - Configuration.hpp (this file) adds reasonable default values for any data missing in Configuration_local_*.hpp.
 - Configuration_adv.hpp holds advanced configuration data, and is not usually required to be changed.
 
Keypad management has been moved out of Utility.* into a new LcdButtons.* files.

A side effect of these changes is that build time is noticeably quicker, and the include files hierarchy in doxygen is much cleaner. I hope that this is a good step along the path to the HAL refectoring.


philharrisathome:
Well, my PR is doing the job and identifying missing pin definitions. Problem is it fails the CI in doing so. I will add some example local configurations tomorrow to allow the CI to complete. These can also act as templates for users.

CTetford:
Just a general comment, but a separate pins file is going to be needed once @andre-stefanov finishes the HAL refactoring. Each supported board type will have it's own file similar to how Marlin/Klipper/Onstep support different boards.

Pin assignments shouldn't be merged with Configuration_local.hpp, but a line should be added within Configuration_local.hpp to select which "pins" file to use based on what board it is. It's the reason when the MKS support was added that a separate file was made within the "boards" directory.

philharrisathome:
Hi @CTetford. This is a good consideration. I've been chatting with @andre-stefanov about the HAL and contributing some material for it. The current HAL approach is indeed for a separate file per board type, however I'm not sure this is going to work well beyond a general ESP32/Atmega platform type id. My reasoning is that there seem to be a lot of OATs out there now and more built every day, and few of these seem to be following a standardized pin configuration. The variation in additional peripherals (displays, keypads, AZ/ALT, etc.) just exacerbates the problem. As a result moving to a pre-defined pin mapping per board type could force a lot of people to rewire their existing hardware and make customization/extension harder, or require them to patch the main OAT code base in multiple places. Neither is ideal.
My preference is for a global user-specified hardware description file that indicates how the hardware resources should be assigned to enabled features. This PR reflects that approach, that a user should only need to modify a single file to bring their OAT up with the features they chose, making it easier to incorporate ongoing changes in the main code base.
Regardless, the changes in this PR support the move towards the HAL by reducing some include cycles in the code base and collating the user-configurable items in one place. We can continue to refactor from there. It could be that the user-centric configuration file I have here ends up being that board-specific definition. This is a design decision that we will need to keep exploring as the HAL work continues.

CTetford:
I completely agree with your approach of having a common file to configure from.

That said, I both agree and disagree with the following:
> My reasoning is that there seem to be a lot of OATs out there now and more built every day, and few of these seem to be following a standardized pin configuration. The variation in additional peripherals (displays, keypads, AZ/ALT, etc.) just exacerbates the problem. As a result moving to a pre-defined pin mapping per board type could force a lot of people to rewire their existing hardware and make customization/extension harder, or require them to patch the main OAT code base in multiple places.

Yes, few people are following a standardized pin mapping - but that's because most people are still using bare Arduinos. Personally I think that approach will go away if we support a couple more RAMPS-based boards (or similar). It really simplifies wiring and software support.

Using off-the-shelf 3D printer boards will essentially force standardized pinmaps for each board type. From a controller perspective, the approach should be no different than similar projects already doing the same thing.

There will still be a generic "Arduino Mega" pin file for those still using bare Arduino boards and they can override the pin definitions in their Configuration_local.hpp as required. This way anyone with an existing OAT would just need to take the time to override the default pin config in a local file, or rewire to match the default - it's their choice.

ClutchplateDude (approved):

ClutchplateDude:
Looks good to me. This includes phil's EEPROM changes, right?

PR #25, opened 2021-01-19
Working ESP32 pin config

trishmapow:
Preliminary working config for ESP32. Serial2 by default is 16/17 so I moved DEC S/D to 26/27. 

I cannot seem to get SoftwareSerial to compile on ESP32. It doesn't seem like it affects anything because the behaviour of ESP32 serial and Arduino SoftwareSerial is the same anyway once it's opened? The ESP32 pins don't get loaded for Mega so there's no need for the extra ifdefs in Config_adv. Apparently with HardwareSerial you can use any pin for serial TX/RX anyway so shouldn't be too hard to add GPS on Serial1.

### DIFF
--- a/Configuration_adv.hpp
+++ b/Configuration_adv.hpp
@@ -331,23 +331,6 @@
 //                                  ////////
 ////////////////////////////////////////////
 
-// Stepper drivers
-#if (RA_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
-  #if defined(ESP32)
-    #define RA_SERIAL_PORT Serial2    // Can be shared with DEC_SERIAL_PORT
-  #elif defined(__AVR_ATmega2560__)
-    // Uses SoftwareSerial
-  #endif
-#endif
-
-#if (DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
-  #if defined(ESP32)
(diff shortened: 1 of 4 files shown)


trishmapow:
Hm that's my exact setup that failed - I just did a clean & build with no issues 🤔

trishmapow:
Hm yeah I understand that but SERIAL_PORT is undefined (only in CI) if I don't include that. Serial2 is just a placeholder, it could be Serial0 or Serial1, I don't know if there's a better/more generic way to refer to it.

philharrisathome:
Hi @trishmapow . The CI tests currently just check that the different builds compile correctly, so the values in _CI are only "valid enough" to pass the build. The pin assignments are however safe invalid values and will be ignored by digitalOutput() etc., so it is safe to deploy on hardware, but it will do nothing.

philharrisathome:
Having said that, the default assignments in all the config files should pass CI tests for both ESP32 and Mega. This was part of PR https://github.com/OpenAstroTech/OpenAstroTracker-Firmware/pull/18 which included valid configuration examples for Mega and ESP32.

trishmapow:
Thanks for the explanation. I think I will close this for now and wait until I/someone else can incorporate GPS, both stepper types and display in one config (your setup on Slack looks good, I will work around that and try to get NEMA working).

PR #27, opened 2021-01-24
Minor fixes for esp32 configurations

philharrisathome:
Minor fixes against the esp32 build:

- Removed #define DISABLED & ENABLED in Constants.hpp due to clash with existing symbols in esp32 IDF. DISABLED & ENABLED don't appear anywhere else in the OAT codebase.
- Removed unnecessary #include "Constants.hpp" in .\boards\* . Constants is already included in the parent Configuration.hpp.
- In Configuration.hpp default BOARD to new BOARD_UNKNOWN value and extend conditional tree at line 253. This is to prevent potential misconfigured hardware in case BOARD_TYPE is undefined: for an esp32 build, if BOARD_TYPE is undefined then Configuration.hpp would apply pin assignments for BOARD_AVR_MEGA2560. Mismatches were possible with other configurations also. With this change no pin assignments are made, however the configuration checks later in Configuration.hpp will provide a suitable message. @CTetford - you may want to review this.

### DIFF
--- a/Configuration.hpp
+++ b/Configuration.hpp
@@ -47,9 +47,10 @@
  * Use default values for any parameters the user didn't provide.
  */
 
-// If board is not defined in local configuration, assume Arduino Mega2560
+// If board is not defined in local configuration, assume BOARD_UNKNOWN
+// Configuration checks below will capture any missing pin assignments.
 #ifndef BOARD
-  #define BOARD BOARD_AVR_MEGA2560
+  #define BOARD BOARD_UNKNOWN
 #endif
 
 // Set to 1 for the northern hemisphere, 0 otherwise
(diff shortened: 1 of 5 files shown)


andre-stefanov:
Mh ... could we sync the changes with hal ideas? Some of these changes already happened in the hal branch

philharrisathome:
I can take a look at making sure these are also merged/addressed into hal branch. It would also be worth merging these into devel for now to fix the warnings on esp32 there. Has all new development switched to hal? I'm still working with the premise that devel feeds release, and that at some point we will switch to hal.

philharrisathome:
Hi @andre-stefanov . Can we get this merged into develop? The build warnings and configuration conflicts fixed by this PR are still present.

ClutchplateDude:
Added remaining changes to PR279

PR #37, opened 2021-02-01
Enforce stricter warnings to build, fix warnings

julianneswinoga:
* Added the following compiler flags to the source tree:
```
-Werror
-Wall
-Wextra
-Wno-unused-parameter
-Wlogical-op
-Wuseless-cast
-Wdouble-promotion
-Wshadow
```
as well as a couple more useful ones that can be un-commented with future refactoring
* Added PUSH_NO_WARNINGS/POP_NO_WARNINGS macro because we don't care about warnings from libraries / library headers
* Changed from `-Os` to `-O2` optimization because we aren't size constrained (yet :grin: )

### DIFF
--- a/pre_build_script.py
+++ b/pre_build_script.py
@@ -1,0 +1,35 @@
+Import("env")
+
+import platform
+
+extra_macros = {
+    # PP indirection needed
+    'DO_PRAGMA_(x)': '_Pragma(#x)',
+    'DO_PRAGMA(x)': 'DO_PRAGMA_(x)',
+
+    # Useful for disabling warnings on system headers that we don't care about
+    'PUSH_NO_WARNINGS': ''.join([
+        'DO_PRAGMA(GCC diagnostic push);',
+        'DO_PRAGMA(GCC diagnostic ignored "-Wconversion");',
+        'DO_PRAGMA(GCC diagnostic ignored "-Wshadow");',
+        'DO_PRAGMA(GCC diagnostic ignored "-Wsign-conversion");',
+        'DO_PRAGMA(GCC diagnostic ignored "-Wsign-compare");',
+        'DO_PRAGMA(GCC diagnostic ignored "-Wignored-qualifiers");',
+        'DO_PRAGMA(GCC diagnostic ignored "-Wuseless-cast");',
+        'DO_PRAGMA(GCC diagnostic ignored "-Wall");',
+        'DO_PRAGMA(GCC diagnostic ignored "-Wextra");',
+        'DO_PRAGMA(GCC diagnostic ignored "-Wpedantic");',
+    ]),
+    'POP_NO_WARNINGS': 'DO_PRAGMA(GCC diagnostic pop);'
+}
+
+def escape_str(s):
+    if 'windows' not in platform.system().lower():
+        # This stuff breaks windows builds
+        if ' ' not in s:
+            s = s.replace('(', '\\(')
+            s = s.replace(')', '\\)')
+    s = s.replace('"', '\\"')
+    return s
+
+env.Append(CPPDEFINES={escape_str(k): escape_str(v) for k, v in extra_macros.items()})
--- a/src/Mount.cpp
+++ b/src/Mount.cpp
@@ -5,11 +5,13 @@
 #include "Mount.hpp"
 #include "Sidereal.hpp"
 
+PUSH_NO_WARNINGS
 #include <AccelStepper.h>
 #if (RA_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART) || (DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART) || \
   (AZ_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART) || (ALT_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
   #include <TMCStepper.h>   // If you get an error here, download the TMCstepper library from "Tools > Manage Libraries"
 #endif
+POP_NO_WARNINGS
 
 //mountstatus
 #define STATUS_PARKED              0B0000000000000000
@@ -44,7 +46,7 @@
 #define SLEW_MASK_ANY     B1111
 
 // Seconds per astronomical day (23h 56m 4.0905s)
-#define SECONDS_PER_DAY 86164.0905
+const float secondsPerDay = 86164.0905f;
 
 const char* formatStringsDEC[] = {
   "",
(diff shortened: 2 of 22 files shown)


julianneswinoga:
@ClutchplateDude @andre-stefanov Are there too many changes in this PR? I would appreciate a review :smile:

ClutchplateDude:
Sorry, Cameron. Trying to find some time to apply this and make sure everything still works as expected.

julianneswinoga:
No worries! I can scale down the changes if you'd like. Reviewing piecemeal might be easier

ClutchplateDude:
So, I applied the PR locally and it does not build. It seems that it does not recognize the PUSH_NO_WARNINGS macro:

<command-line>:0:16: note: in expansion of macro 'DO_PRAGMA_'
<command-line>:0:253: note: in expansion of macro 'DO_PRAGMA'
src\inc/Globals.hpp:3:1: note: in expansion of macro 'PUSH_NO_WARNINGS'
 PUSH_NO_WARNINGS
 ^
<command-line>:0:26: error: stray '#' in program
<command-line>:0:16: note: in expansion of macro 'DO_PRAGMA_'
<command-line>:0:17: note: in expansion of macro 'DO_PRAGMA'
src\inc/Globals.hpp:6:1: note: in expansion of macro 'POP_NO_WARNINGS'
 POP_NO_WARNINGS
 ^
src\a_inits.hpp:15:1: error: stray '\' in program

julianneswinoga:
Interesting, can you attach your full build log with verbose mode on?

ClutchplateDude:
Are you on our Slack channel? This isn't the best way to discuss here....

julianneswinoga:
Yep just my name there too, you can DM if you'd like. I just like the transparency of a public channel, easier searching if someone has the same problem haha

ClutchplateDude:
I did, lol. Kind of. Not DM. Public channel.

julianneswinoga:
@ClutchplateDude Building on Windows should be fixed now :smile:

julianneswinoga:
@ClutchplateDude @andre-stefanov Any more comments on this PR? I've tested the builds under Linux and Windows, with both platfomio and Arduino IDE

src/Mount.cpp
```diff
@@ -5,10 +5,14 @@
...
 
+PUSH_NO_WARNINGS
 #include <AccelStepper.h>
+POP_NO_WARNINGS
 #if (RA_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART) || (DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART) || \
   (AZ_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART) || (ALT_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
+  PUSH_NO_WARNINGS
   #include <TMCStepper.h>   // If you get an error here, download the TMCstepper library from "Tools > Manage Libraries"
+  POP_NO_WARNINGS
 #endif
```
andre-stefanov: in theory we dont need to do thing for each include, we could just wrap all the external lib/framework includes (as block) with this macro

src/Utility.hpp
```diff
@@ -131,6 +131,20 @@ void logv(int levelFlags, String input, ...);
 
 #endif // DEBUG_LEVEL>0
 
+// For some reason arduino just defines all float functions to be as double
```
andre-stefanov, line 134: did you test this also on esp32? on AVR float and double are actually the same (32bit). but what if it is different for esp32? could we break some external libs this way?

pre_build_script.py
```diff
@@ -0,0 +1,35 @@
...
+
+import platform
+
+extra_macros = {
+    # PP indirection needed
+    'DO_PRAGMA_(x)': '_Pragma(#x)',
+    'DO_PRAGMA(x)': 'DO_PRAGMA_(x)',
+
+    # Useful for disabling warnings on system headers that we don't care about
+    'PUSH_NO_WARNINGS': ''.join([
```
andre-stefanov, line 11: just a small one ... what if we rename it to "DISABLE_WARNINGS" and "ENABLE_WARNINGS"? or is there some convention to name it push and pop?

andre-stefanov (changes requested):

src/Mount.cpp
julianneswinoga: Yeah I know, I felt this was a bit cleaner since there's some preprocessor control blocks which break the includes up. I can move them to a single `PUSH`/`POP` if you think that's cleaner

pre_build_script.py
julianneswinoga: We can if you'd like! I was just following the convention from the `#pragma GCC diagnostic push`/`#pragma GCC diagnostic pop` that the macros use (and pushing/popping is a pretty common term for stuff like this)

src/Utility.hpp
julianneswinoga: I haven't tested on esp32 (I don't have the hardware) but this really shouldn't break anything. Worst case scenario is that a `double` is implicitly demoted to `float` but that can be fixed with adding `-Wconversion` in a later PR (decent amount of code needs to be fixed in order to turn that one on haha)

pre_build_script.py
andre-stefanov: i think it depends on how it works. if you e.g. push 2 times and need at least 2 times pop (stack logic), then we should keep PUSH/POP naming. otherwise enable/disable is more correct

src/Utility.hpp
andre-stefanov: Yeah good idea. would be a bummer if some esp32 library really needs double precision and we break it this way.

src/Mount.cpp
andre-stefanov: I think it would be easier to read and also a clean simple convention: "wrap all external library includes with NO_WARNINGS". Probably an even better solution would be to disable the checks for the external libs per compiler flags in pio ... but I dont know (yet) if it is possible.

julianneswinoga: Yeah I initially tried to have platformio use `-isystem` when including external headers so that GCC suppresses all warnings ([system header documentation](https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html)) but platformio wasn't playing nice... I really delved deep into how platformio works (i.e. [hacking around with SCons](https://docs.platformio.org/en/latest/projectconf/advanced_scripting.html)) but I couldn't get it to do what I wanted. The macros are a pretty ugly solution but I think it's the only maintainable way to do what we want.

pre_build_script.py
julianneswinoga: Ah I see what you mean, GCC does keep a stack so the `PUSH`/`POP` naming is technically correct ([GCC docs](https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html))

andre-stefanov (approved):

PR #44, opened 2021-02-06
Reset init

ClutchplateDude:


src/Mount.cpp
```diff
@@ -121,6 +127,10 @@ Mount::Mount(LcdMenu* lcdMenu) :
 void Mount::clearConfiguration()
 {
   EEPROMStore::clearConfiguration();
+  _stepsPerRADegree = RA_STEPS_PER_DEGREE;
+  _stepsPerDECDegree = DEC_STEPS_PER_DEGREE;
```
andre-stefanov: Why is it not part of `initializeVariables()`? Also setting these values here is probably not needed because they get directly overriden in`readPersistentData()` which gets called by `readConfiguration()`

andre-stefanov (changes requested):

src/Mount.cpp
andre-stefanov: I would propose to move these assignments to `initializeVariables()` and ignore the fact that they get overriden from eeprom. This way it would be safe to call `initializeVariables()` even without reading eeprom afterwards.

andre-stefanov (approved):

PR #52, opened 2021-02-08
Fix Arduino IDE build

### LINKED ISSUE #51: Build broken for Arduino IDE
Using `git bisect` and `arduino-cli` I determined that the build was broken in commit 26c2a22530d6e538d6adeb5e472311a997402781

julianneswinoga:
Fixes #51 
The Arduino IDE expects all source files to be in the src/ directory,
else it will not be able to find and include them in the build process.

This kinda sucks, since people with `Configuration_local.hpp` files will have to move them into `src/` else they wont be included... Do we have mechanisms to tell people about this change?
@ClutchplateDude @andre-stefanov

### DIFF
--- a/boards/AVR_MEGA2560/pins_MEGA2560.hpp
+++ b/boards/AVR_MEGA2560/pins_MEGA2560.hpp
@@ -4,8 +4,6 @@
 
 #pragma once
 
-#include "Constants.hpp"
-
 // DRIVER_TYPE_ULN2003 requires 4 digital outputs in Arduino pin numbering
 #ifndef RA_IN1_PIN
   #define RA_IN1_PIN  22
--- a/boards/AVR_MKS_GEN_L_V21/pins_MKS_GEN_L_V21.h
+++ b/boards/AVR_MKS_GEN_L_V21/pins_MKS_GEN_L_V21.h
@@ -4,8 +4,6 @@
 
 #pragma once
 
-#include "Constants.hpp"
-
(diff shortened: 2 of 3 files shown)


andre-stefanov:
> Fixes #51
> The Arduino IDE expects all source files to be in the src/ directory,
> else it will not be able to find and include them in the build process.
> 
> This kinda sucks, since people with `Configuration_local.hpp` files will have to move them into `src/` else they wont be included... Do we have mechanisms to tell people about this change?
> @ClutchplateDude @andre-stefanov

This is not true. See https://arduino.github.io/arduino-cli/sketch-build-process/#pre-processing Arduino IDE does not handle src folder in any special way (PlatformIO does if not configured differently). 
I did not have a chance to look deeper into this issue, but current folder structure was intentional to reduce amount of tabs on Arduino IDE for the user.

julianneswinoga:
Interesting, the documentation I found all referenced that only the `src` folder is included recursively: https://arduino.stackexchange.com/a/54655
I'll investigate that doc link (could not find that type of documentation earlier for whatever reason) but qualitatively the build would only work when I moved all includes into the `src` directory.

julianneswinoga:
Turns out I'm dumb and the build only needed a minor change to a header include. Big thanks to the arduino docs for being so confusing! :grinning:

CTetford:
Rather than fixing the relative path, could you remove the #include on all the pins files (not just the mega)? It isn't actually used in any of them.

CTetford (approved):

ClutchplateDude (approved):

PR #53, opened 2021-02-10
Cleaned up and improved configs and matrix build

andre-stefanov:
* refactored (cleaned up) config file structure
* complete matrix build rewrite


andre-stefanov:
there are CI adaption missing

PR #54, opened 2021-02-10
Build refactoring

andre-stefanov:
- refactoring configurations
- moved validation into a separate header
- complete matrix build script refactoring
- added PIO environments for mks boards
- changed ci workflow to use matrix_build script (maybe some adaptions will be still needed, lets see how it works)
- changed default envs to none (commented out). if the user wants to build only for one environment/board, he can select it easily in VSC


Configuration.hpp
```diff
@@ -32,22 +32,22 @@
...
-  #define BOARD BOARD_AVR_MEGA2560
+  // #define BOARD BOARD_AVR_MEGA2560
 #endif
 
+#ifndef BOARD
+  #error You have to specify the board
+#endif
+
+// Include the user-specific local configuration
+#include "LocalConfiguration.hpp"
```
ClutchplateDude, line 50: Yet another configuration renaming... users are going to love us.
[Later] Ahh, I see this is just an extra level.

Configuration.hpp
```diff
@@ -32,22 +32,22 @@
...
-#endif
-
 /**
  * Use default values for any parameters the user didn't provide.
  */
 
 // If board is not defined in local configuration, assume Arduino Mega2560
 #ifndef BOARD
-  #define BOARD BOARD_AVR_MEGA2560
+  // #define BOARD BOARD_AVR_MEGA2560
```
ClutchplateDude, line 42: May as well delete these three lines if you're going to error out right after anyway.

ClutchplateDude (changes requested):

Configuration.hpp
andre-stefanov: Yeah this is just to isolate all the ifdefs of the main configuration file and keep it cleaner

Configuration.hpp
andre-stefanov: My idea was to provide an example here this way (in case people use Arduino IDE). platformio sets the board automatically

andre-stefanov: but i see now the comment does not match now :)

andre-stefanov: I have adapted the code comment. In general we need this definition because Arduino IDE is not able to set it. But with new comment it should be hopefully clear when and how to set it.

ClutchplateDude (approved):

PR #55, opened 2021-02-10
change default dec step and dir pins on esp32 to free up serial2

braingram:
I ran into an issue trying to use the default pins on a esp32 build with tmc2209 stepper drivers configured via uart. The default step and direction pins for the dec axis on the esp32 platform are pins 16 and 17. These are also the RX and TX pins for Serial2 which is used for uart configuration of the tmc2209 stepper drivers.

This pull request changes the default step and direction pins to some other free pins on the esp32 to hopefully avoid this conflict.


trishmapow:
Hm yes this works but still blocks the I2C pins for display and joystick. I think it would be a good idea to change all the pins in one go to match @philharrisathome 's configuration [here](https://github.com/philharrisathome/OpenAstroTracker-Hardware-Local/blob/main/OAT%20ESP32%20Ports%20v2.pdf). I can now confirm that those pins also work for TMC2209. I can make a PR for this unless someone else wants to do it.

PR #66, opened 2021-02-27
Fixup/cs/turn defines into enums

julianneswinoga:
This allows for better type checking vs raw integers. Coming from C, hard type checks on enums is a real treat! Note the inclusion of `default:` cases where previously it was not handled.


src/LcdButtons.cpp
```diff
@@ -10,11 +10,11 @@ LcdButtons::LcdButtons(byte pin, LcdMenu* lcdMenu) {
     _analogPin = pin;
     _lastKeyChange = 0;
 
-    _newKey = -1;
-    _lastNewKey = -2;
+    _newKey = btnNONE;
```
julianneswinoga, line 13: I don't think this is a functional change, but we could add `btnINVALID` if this could cause issues.

src/c65_startup.hpp
```diff
@@ -15,22 +15,24 @@
...
-#define StartupWaitForRollCompletion 3
-#define StartupRollConfirmed 4
-#define StartupSetHATime 10
-#define StartupWaitForHACompletion 15
-#define StartupHAConfirmed 20
-#define StartupWaitForPoleCompletion 25
-#define StartupPoleConfirmed 30
-#define StartupCompleted 35
+enum startupState_t {
+    StartupIsInHomePosition = 1,
```
julianneswinoga: @andre-stefanov Is there meaning to the numbering scheme here? If not we could just drop the assignment in the enumeration

ClutchplateDude: You can drop the numbers

src/LcdButtons.cpp
```diff
@@ -10,11 +10,11 @@ LcdButtons::LcdButtons(byte pin, LcdMenu* lcdMenu) {
...
+    _lastNewKey = btnNONE;
 
-    _currentKey = -1;
-    _lastKey = -2;
+    _currentKey = btnNONE;
+    _lastKey = btnNONE;
```
ClutchplateDude, line 17: I need to check about this, but I think last is initialized different from current so that the first comparison fails (if (current!=last))
I think this would miss the very first key press after boot up.

ClutchplateDude (changes requested):

ClutchplateDude (approved):

julianneswinoga:
@ClutchplateDude Don't merge yet please, I'd like to clean up the commit history

Configuration.hpp
```diff
@@ -258,5 +258,8 @@
   #include "boards/AVR_MKS_GEN_L_V21/pins_MKS_GEN_L_V21.h"
 #endif
 
+#define USE_DUMMY_EEPROM false
+#define BUFFER_LOGS false
```
andre-stefanov: i would not place these into the Configuration.hpp but rather Configuration_adv.hpp because this is not one of the basic configs most users need for their OATs

ClutchplateDude:
> @ClutchplateDude Don't merge yet please, I'd like to clean up the commit history

I can squash and merge if you like,

julianneswinoga:
> > @ClutchplateDude Don't merge yet please, I'd like to clean up the commit history
> 
> I can squash and merge if you like,

No I like to keep the changes in separate commits (esp for simple changes like this). Good to merge whenever :)

andre-stefanov (changes requested):

andre-stefanov (approved):

PR #68, opened 2021-03-01
Fix RA UART rms current initialization

CTetford:
Currently when the RA driver is initialized, the TMCStepper rms_current() function is called to set the RMS current for the motor.

Immediately afterward both the ihold() and irun() functions are called, which overrides what was set by rms_current().

irun(31) is called, setting the RMS run current to maximum (with the Vref potentiometer as the only method to control RA current).

The DEC/AZ/ALT driver initializations appear to be correct.

### DIFF
--- a/src/Mount.cpp
+++ b/src/Mount.cpp
@@ -367,9 +367,8 @@ bool Mount::connectToDriver( TMC2209Stepper* driver, const char *driverKind ) {
     #if UART_CONNECTION_TEST == 1
       connectToDriver( _driverRA, "RA" );
     #endif
-    _driverRA->toff(4);
     _driverRA->blank_time(24);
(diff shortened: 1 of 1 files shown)


CTetford:
NOTE: Merging this PR may dramatically reduce the effective RMS output of their existing configs, since they have always been at the max setting. There should be a warning with this update that the RMS current values may need adjustment.

andre-stefanov (approved):

CTetford:
> NOTE: Merging this PR may dramatically reduce the effective RMS output of their existing configs, since they have always been at the max setting. There should be a warning with this update that the RMS current values may need adjustment.

@andre-stefanov @ClutchplateDude 

Just bringing this to everyone's attention in case it was missed......

PR #72, opened 2021-03-09
Fixup/cs/turn more defines into enums

julianneswinoga:
Biggest addition in this PR is the `MappedDict` class which allows the simple lookup of one value to another value. It's meant to act like `std::map` ([docs](https://en.cppreference.com/w/cpp/container/map)) because we do not really have the STL on embedded platforms.
The `bool get(KeyType query, ValueType *rtnValPtr)` method semantics (return code and pointer for output value) are such because we do not have exceptions on embedded platforms either. If there were exceptions, we could simply return the value if it is in the map else throw an exception but unfortunately we do not.
Using `MappedDict` in certain places where a `switch` or `if/else` chain would be used makes the intention of the developer a lot more clear. :smile: Let me know your thoughts and if you think there is a cleaner way to go about the implementation!

The latest commit (3638d780333) does deeper refactoring of `c75_menuCTRL.hpp`, it's not necessary and if wanted I will drop it, however I do think it is a bit of a cleaner implementation!

Additionally, I believe that the refactoring of `processKeyStateChanges()` (now `controlManualSlew()`) in 913057f9fea fixes a bug where pressing alternating buttons quickly would lead to locking up the logic and effectively locking out any further keypresses.

### DIFF
--- a/src/LcdMenu.cpp
+++ b/src/LcdMenu.cpp
@@ -1,6 +1,7 @@
 #include "inc/Globals.hpp"
 #include "../Configuration.hpp"
 #include "Utility.hpp"
+#include "MappedDict.hpp"
 #include "EPROMStore.hpp"
 #include "LcdMenu.hpp"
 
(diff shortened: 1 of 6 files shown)


src/MappedDict.hpp
```diff
@@ -0,0 +1,43 @@
...
+            : _dictPtr(dictPtr), _dictSize(dictSize) {
+    }
+
+    /**
+     * Try to retrieve a value from the mapping
+     * @param[in] query Key to look for in the dictionary
+     * @param[out] rtnValPtr Pointer to store the return value in, if successful
+     * @return true if the query was found in the dictionary, false otherwise
+     */
+    bool get(KeyType query, ValueType *rtnValPtr) {
```
ClutchplateDude: I'd like to see this named tryGet(..) That's more in line what im used to with this kind of interface

src/LcdMenu.cpp
```diff
@@ -343,9 +312,9 @@ void LcdMenu::printMenu(String line)
 
     _lcd.setCursor(_activeCol, _charHeightRows*_activeRow);
     int spaces = _columns - line.length();
-    for (unsigned int i = 0; i < line.length(); i++)
+    for (char i : line)
```
ClutchplateDude, line 315: What is this voodoo magic syntax? I must have missed a C++ standard along the way, lol. Nice.

src/MappedDict.hpp
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+/**
+ * @brief A simple class to deal with 1-1 value mapping, a la python dictionaries
+ * @details Operations operate on pointers so that this class has no data storage,
+ * data should be set up by the caller and passed in to the constructor.
+ * @tparam KeyType Type that should be the 'key'
+ * @tparam ValueType Type that should be the 'value'
+ */
+template<class KeyType, class ValueType>
```
ClutchplateDude, line 10: Uh-oh, we have the first template in the project... it's all downhill from here.... :-)

ClutchplateDude (changes requested):
This is great, love it! Thanks for adding this stuff.

src/LcdMenu.cpp
julianneswinoga: Yep! Range based for loops were added in C++11 ([docs](https://en.cppreference.com/w/cpp/language/range-for)) and it can really clean up simple string loops like this :grin:

src/MappedDict.hpp
julianneswinoga: Haha yeah I thought that since this is a pretty simple and easy to follow implementation of templates it would be a good first introduction to them in this project :)

src/LcdMenu.cpp
```diff
@@ -239,83 +240,51 @@ void LcdMenu::updateDisplay()
...
-  {
-    _lcd.print(ch);
-  }
+    MappedDict<char, specialChar_t>::DictEntry_t lookupTable[] = {
+            {'>',  _rightArrow},
+            {'<',  _leftArrow},
+            {'^',  _upArrow},
+            {'~',  _downArrow},
+            {'@',  _degrees},
+            {'\'', _minutes},
```
julianneswinoga, line 276: I also noticed that we don't really need this minutes custom bitmap? It's just an apostrophe, and the SSD1306 implementation doesn't have this mapped to a special font.

ClutchplateDude: The standard apostrophe on the 16x2 LCD is ugly, it's an upside down comma. so a square with a tail. This is a 3-pixel line like the double quotes use. Call me picky.

src/MappedDict.hpp
ClutchplateDude: Templates are great! As long as used sparingly. Once we get to templates of templates or metatemplates, I'm checking out :-)

src/LcdMenu.cpp
ClutchplateDude: Also, the support for this LCD screen is a user contribution.

ClutchplateDude (approved):

PR #73, opened 2021-03-11
Bugfix/cs/fix errors when compiling with debug

julianneswinoga:
Turning `DEBUG_ANY` on caused the compiler to emit a warning (error) about variable shadowing. We should really have a CI test that enables debug output.


ClutchplateDude:
I already fixed this one, albeit a little differently....

julianneswinoga:
> I already fixed this one, albeit a little differently....

Where did you fix this? I don't see any new commits on develop

ClutchplateDude:
Yeah, not committed yet.

PR #74, opened 2021-03-11
Enable LCD brightness again for keypad displays

julianneswinoga:
This is an interesting one! Turns out the crashes reported earlier were because of manufacturing defects which cause the brightness pin to short to ground. When the LCD brightness pin is turned on it consumes too much power and thus cause a brownout.

This fix (helped largely by finding https://forum.arduino.cc/index.php?topic=96747.0) will 'test' the LCD display at startup and store a value if it exhibits the electrical fault. We can then check that value at runtime to:
- limit the faulty LCD to full brightness or completely off
- change how we control said brightness (either `pinMode(INPUT);` for ON or `pinMode(OUTPUT);` for OFF)

### DIFF
--- a/src/LcdMenu.cpp
+++ b/src/LcdMenu.cpp
@@ -41,14 +41,19 @@ void LcdMenu::startup()
 
   #if DISPLAY_TYPE == DISPLAY_TYPE_LCD_KEYPAD
     _lcd.begin(_cols, _rows);
+    #if defined(LCD_BRIGHTNESS_PIN)
+      _lcdBadHw = testIfLcdIsBad();
+    #endif
   #elif DISPLAY_TYPE == DISPLAY_TYPE_LCD_KEYPAD_I2C_MCP23008 || DISPLAY_TYPE == DISPLAY_TYPE_LCD_KEYPAD_I2C_MCP23017
     _lcd.begin(_cols, _rows);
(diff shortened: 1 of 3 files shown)


ClutchplateDude (changes requested):
So I seem to have a bad one :-). Since I won't be able to see anything if the brightness is 0, I wonder if we shouldn't just disable this menu if the LCD is bad.... Or change it from "LCD Brightness" to LCD On/Off".... thoughts?

julianneswinoga:
> So I seem to have a bad one :-). Since I won't be able to see anything if the brightness is 0, I wonder if we shouldn't just disable this menu if the LCD is bad.... Or change it from "LCD Brightness" to LCD On/Off".... thoughts?

I want the ability to turn on/off the backlight, one it saves battery and two it keeps your night vision :)
(It's still possible to read the display with an external light pointed at it, so you can turn it back on)
I thought about changing the control to "On/Off" but:
- 0/1 is easy to understand by a user, it's pretty obvious what it does just by clicking up/down :)
- The LCD interface would need a bit more refactoring to implement this cleanly, perhaps we just keep this feature in mind when developing the new HAL @andre-stefanov

ClutchplateDude (approved):
Looks good

PR #76, opened 2021-03-11
Fix TMC2209 UART / rms_current settings + Test Method for UART Tx-Only

CTetford:
Fixes the way rms_current sent to the TMC2209 to be more accurate and consistent.

Disables i_scale_analog (aka "USE_VREF") by default so that results are more consistent between users. rms_current will no longer depend on Vref, unless specifically configured by the user.

Attempts to retrieve settings from the TMC2209 and outputs them to log to allow verification that settings are applied properly.

Changes UART baud rate to 19200 rather than the default of 115200 - the Rx channel was too unreliable and repeatedly had CRC errors at the higher baud rate.

Adds a test method for visual confirmation of Tx-only UART connectivity. At boot (before tracking starts), each axis will rotate slightly in each direction driven purely by UART commands. If the user sees the movement, they know UART Tx is functioning


ConfigurationValidation.hpp
```diff
@@ -302,14 +302,67 @@
...
-#if (RA_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART) && (RA_RMSCURRENT > 2000)
-  #error "Do you really want to set the RMS motorcurrent above 2 Ampere? Thats almost 3A peak! Delete this error if you know what youre doing" 
-#endif
+#if (AZIMUTH_ALTITUDE_MOTORS == 1)
+  #if (AZ_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
+    #if (AZ_MOTOR_CURRENT_RATING > 1700)
+      #error "The TMC2209 driver is only rated up to 1.7A output. Delete this error if you know what youre doing" 
+    #endif
+    #if (AZ_MOTOR_CURRENT_RATING == 0)
+      #error "RA current rating/setting is not configured. Enter the current rating of your motor"
```
ClutchplateDude: RA should be AZ

ConfigurationValidation.hpp
```diff
@@ -302,14 +302,67 @@
...
+    #else
+        #error "AZ_OPERATING_CURRENT_SETTING not defined."
+    #endif
+  #endif
+  #if (ALT_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
+    #if (ALT_MOTOR_CURRENT_RATING > 1700)
+      #error "The TMC2209 driver is only rated up to 1.7A output. Delete this error if you know what youre doing" 
+    #endif
+    #if (ALT_MOTOR_CURRENT_RATING == 0)
+      #error "RA current rating/setting is not configured. Enter the current rating of your motor"
```
ClutchplateDude: RA -> AZ

Configuration_adv.hpp
```diff
@@ -300,13 +307,20 @@
...
-  #define ALT_RMSCURRENT 1000   // RMS current in mA. Warning: Peak current will be 1.414 times higher!! Do not exceed your steppers max current!
-
+  #if (AZ_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART) || (ALT_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
+    #define AZ_RMSCURRENT AZ_MOTOR_CURRENT_RATING * (AZ_OPERATING_CURRENT_SETTING / 100.0f) / 1.414f
+    #define ALT_RMSCURRENT ALT_MOTOR_CURRENT_RATING * (ALT_OPERATING_CURRENT_SETTING / 100.0f) / 1.414f
+    
+    #define AZ_AUDIO_FEEDBACK 0 // of the stepper coils. Use this to verify that UART is working properly. 
+    #define ALT_AUDIO_FEEDBACK 0 // of the stepper coils. Use this to verify that UART is working properly.
+    
+    #define AZ_STALL_VALUE 10    // adjust this value if the RA autohoming sequence often false triggers, or triggers too late
```
ClutchplateDude, line 317: Comment talks about homing....

Configuration_adv.hpp
```diff
@@ -300,13 +307,20 @@
...
-  #define AZ_STALL_VALUE 10    // adjust this value if the RA autohoming sequence often false triggers, or triggers too late
-  #define AZ_RMSCURRENT 1000   // RMS current in mA. Warning: Peak current will be 1.414 times higher!! Do not exceed your steppers max current!
-  #define ALT_STALL_VALUE 10    // adjust this value if the RA autohoming sequence often false triggers, or triggers too late
-  #define ALT_RMSCURRENT 1000   // RMS current in mA. Warning: Peak current will be 1.414 times higher!! Do not exceed your steppers max current!
-
+  #if (AZ_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART) || (ALT_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
+    #define AZ_RMSCURRENT AZ_MOTOR_CURRENT_RATING * (AZ_OPERATING_CURRENT_SETTING / 100.0f) / 1.414f
+    #define ALT_RMSCURRENT ALT_MOTOR_CURRENT_RATING * (ALT_OPERATING_CURRENT_SETTING / 100.0f) / 1.414f
+    
+    #define AZ_AUDIO_FEEDBACK 0 // of the stepper coils. Use this to verify that UART is working properly.
```
ClutchplateDude: Comment is from copy paste. needs to be updated

Configuration_sample_local.hpp
```diff
@@ -43,8 +43,12 @@
 
 // TMC2209 UART settings
 // These settings work only with TMC2209 in UART connection (single wire to TX)
-#define RA_RMSCURRENT 1200       // RMS current in mA. Warning: Peak current will be 1.414 times higher!! Do not exceed your steppers max current!
-#define DEC_RMSCURRENT 1000      // RMS current in mA. Warning: Peak current will be 1.414 times higher!! Do not exceed your steppers max current!
+#define RA_MOTOR_CURRENT_RATING 0       // Current rating of RA motor in mA
```
ClutchplateDude, line 46: Why is this zero? If we are expecting users to set this, we should require it in the config_local and #ifndef error here

ClutchplateDude (changes requested):

Configuration_adv.hpp
CTetford: I'm not sure when that got added - it wasn't added in this PR, just shifted around (previously line 305)

Configuration_adv.hpp
CTetford: Fixed. I'm tempted to delete the audio feedback functionality altogether now that there's an alternative. Let me know what you think.

ConfigurationValidation.hpp
CTetford: Fixed

ConfigurationValidation.hpp
CTetford: Fixed

Configuration_sample_local.hpp
CTetford: The user is required to edit this according to their specific motor rating. It's required when using UART, so it should be defined as 'something' by default - zero works. ConfigurationValidation will throw an error if it remains at zero or is >1700 (the maximum output for a TMC2209).

I've added an #ifndef error to ConfigurationValidation in case someone removes it from the local config.

ConfigurationValidation.hpp
```diff
@@ -303,11 +303,15 @@
...
-    #error "The TMC2209 driver is only rated up to 1.7A output. Delete this error if you know what youre doing" 
-  #endif
-  #if (DEC_MOTOR_CURRENT_RATING == 0)
-    #error "DEC current rating/setting is not configured. Enter the current rating of your motor" 
+  #if defined(DEC_MOTOR_CURRENT_RATING)
+    #if (DEC_MOTOR_CURRENT_RATING > 1700)
+      #error "The TMC2209 driver is only rated up to 1.7A output. Delete this error if you know what youre doing" 
+    #endif
+    #if (DEC_MOTOR_CURRENT_RATING == 0)
+      #error "DEC current rating/setting is not configured. Enter the current rating of your motor"
```
ClutchplateDude: This message is not going to allow users to recover easily. The second sentence should read something like:

Please define the current rating of your motor in you local configuration file using the DEC_MOTOR_CURRENT_RATING keyword.

ClutchplateDude: Same for line 314

ClutchplateDude (changes requested):

ConfigurationValidation.hpp
CTetford: Fixed

ClutchplateDude (approved):
Looks good

CTetford:
@ClutchplateDude - I'd like to give @andre-stefanov a chance to do some RMS testing before this gets merged. Based on what I can do, everything seems accurate but he's got better tools.

PR #77, opened 2021-03-14
Add debugger support to ATmega2560 family via avr-stub

julianneswinoga:
This adds support for the platformio debug interface on ATmega2560 platforms via the `avr-stub` library.

Glorious debugging session:
debugging_screenshot

Note the two extra scripts required to use the `avr-stub` library. We basically need to patch `WInterrupts.c` so as to not use INT7 so that the debug library can use it. Another limitation is that Serial output cannot be used when debugging. These are partially explained in the updates to the `README.md`

I also did minor cleanup of `platformio.ini` to use the `extends` keyword ([docs](https://docs.platformio.org/en/latest/projectconf/section_env_advanced.html#extends)) on mega2560 based boards.


julianneswinoga:
Ping @andre-stefanov, have any quick thoughts on this addition?

Configuration_adv.hpp
```diff
@@ -445,4 +447,26 @@
...
+#if defined(OAT_DEBUG_BUILD)
+  // AVR based boards have numbers < 1000
+  #if BOARD < 1000
+    /*
+     * Debugging on the mega2560 using avr-stub dissallows application-code from
+     * using the normal (USB) serial port
+     */
+    // Disable debug output
+    #if defined(DEBUG_LEVEL)
+      #undef DEBUG_LEVEL
```
ClutchplateDude, line 478: So we don't want logging while debugging? Any particular reason for this? Is Serial unavailable?

[Later]
Ahh, I see that Serial is indeed not available.... which begs the question, how do you make it do the things you want to debug?

Configuration_adv.hpp
```diff
@@ -445,4 +447,26 @@
...
+    #endif
+    #define DEBUG_LEVEL (DEBUG_NONE)
+
+    // Disable serial control
+    #if defined(SUPPORT_SERIAL_CONTROL)
+      #undef SUPPORT_SERIAL_CONTROL
+    #endif
+    #define SUPPORT_SERIAL_CONTROL false
```
ClutchplateDude: We use 0 and 1 everywhere for defines, can we stick with that, please?

ClutchplateDude (changes requested):

Configuration_adv.hpp
julianneswinoga: Yeah I can change them to match the old defines, I just find it a lot more readable to use true booleans when you're looking through the logic

Configuration_adv.hpp
julianneswinoga: Yeah `avr-stub` needs total control of the serial port. There is `debug_message()` which prints messages to the GDB console (haven't really looked into it too much though) and it's also possible to still use the LCD to navigate. I could look into substituting our serial interface with the GDB console in the future

Configuration_adv.hpp
ClutchplateDude: I hear you. I think we decided on numbers because there's just always such confusion around using true, True or TRUE and unfortunately they are three distinct values to the preprocessor. The numbers are not ambiguous :-)

Configuration_adv.hpp
ClutchplateDude: If the LCD works, I think it's fine. The Meade layer is not that complex, I think the light handed DEBUG_LEVEL can handle that.

Configuration_adv.hpp
julianneswinoga: Partially agree, but I just want to put it out there that `false` and `true` are defined in the standard as 0/1 respectively (["If the source type is `bool`, the value `false` is converted to `0` and the value `true` is converted to the value `1` of the destination type"](https://en.cppreference.com/w/cpp/language/implicit_conversion#Integral_conversions))

I'm gonna change it, but I'll always try and convince people for easier to read software that is strictly standards-compliant 😄

ClutchplateDude (approved):
LGTM!

PR #81, opened 2021-03-22
Serial and UART test fixes

CTetford:


ClutchplateDude:
Can we get a description of what these fixes do and why only the AZ and ALT systems are getting them?

CTetford:
I'm playing catch-up. @intercipere already implemented this to the RA/DEC driver configs. UART microstepping doesn't work without it when using the TMCStepper function "beginSerial" rather than "begin"

The other change is self explanatory - so we can add the audio test to Configuration_local rather than modifiying Configuration_adv.

ClutchplateDude (approved):

PR #83, opened 2021-03-24
Improved UART TX test

CTetford:
Add configurable option to UART TX test to specify approximate degrees of test rotation


ClutchplateDude (dismissed):

andre-stefanov:
mh it is possible to enable both test types at the moment. we should use an enumeration for the test type (e.g. values defined in Constants.hpp like for other hardware stuff).

andre-stefanov:
5° on DEC is good visible but on RA i had to really pay close attention when not seeing the oat from the front (which is the case for some builds).

andre-stefanov (changes requested):
mh it is possible to enable both test types at the moment. we should use an enumeration for the test type (e.g. values defined in Constants.hpp like for other hardware stuff).

src/Mount.cpp
```diff
@@ -3691,31 +3691,46 @@ DayTime Mount::calculateHa()
...
+    _driverRA->microsteps(RA_TRACKING_MICROSTEPPING == 1 ? 0 : RA_TRACKING_MICROSTEPPING);
 }
     #endif
 
     #if DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART
 void Mount::testDEC_UART_TX()
 {
-    int _speed    = 1000;  //microsteps per driver clock tick
-    int _duration = 1000;  //Duration to run in ms
+    int _speed = (DEC_STEPPER_SPEED / 2)
```
julianneswinoga: Local variables shouldn't start with an underscore

julianneswinoga: You should add `const` to these local variables

src/Mount.cpp
```diff
@@ -3691,31 +3691,46 @@ DayTime Mount::calculateHa()
...
+                 / 0.715255737f;  //microsteps per driver clock tick (From TMC2209 datasheet: v[Hz] (microstep/s) = VACTUAL[2209] * 0.715Hz)
```
julianneswinoga: Could you move these comments to the line above the variable? That way you don't need to add line breaks in the middle of the expression

src/Mount.cpp
```diff
@@ -3691,31 +3691,46 @@ DayTime Mount::calculateHa()
...
     testUART_vactual(_driverDEC, _speed, _duration);
+    LOGV2(DEBUG_STEPPERS, F("[STEPPERS]: uartTest: Switching DEC driver to microsteps(%d) after UART test"), DEC_GUIDE_MICROSTEPPING);
+    _driverDEC->microsteps(DEC_GUIDE_MICROSTEPPING == 1 ? 0 : DEC_GUIDE_MICROSTEPPING);
 }
     #endif
 
     #if RA_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART || DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART
 void Mount::testUART_vactual(TMC2209Stepper *driver, int _speed, int _duration)
-{
+{  //Runs motor at specified speed for specified duration in each dirction, allowing 0.5s to stop motion after each move
```
julianneswinoga: This comment should go above the function, ideally in docblock format:
```C++
/**
 * Runs motor at specified speed for specified duration in each direction, allowing 0.5s to stop motion after each move
 * @param[in,out] driver The stepper driver instance
 * @param[in] speed The speed to run the stepper motor at (in deg/s? idk)
 * @param[in] duration The amount of time to turn the stepper, in seconds
 */
void Mount::testUART_vactual(TMC2209Stepper *driver, int speed, int duration)
{
```

julianneswinoga (changes requested):

andre-stefanov (approved):

andre-stefanov (dismissed):
Oops

src/Mount.cpp
CTetford: Done

src/Mount.cpp
CTetford: Done

src/Mount.cpp
CTetford: Done

src/Mount.cpp
julianneswinoga: ? I don't see it changed haha

CTetford: Sorry - my bad, I skipped one lol

src/Mount.cpp
```diff
@@ -3691,31 +3691,52 @@ DayTime Mount::calculateHa()
...
+    LOGV2(DEBUG_STEPPERS, F("[STEPPERS]: uartTest: Switching DEC driver to microsteps(%d) after UART test"), DEC_GUIDE_MICROSTEPPING);
+    _driverDEC->microsteps(DEC_GUIDE_MICROSTEPPING == 1 ? 0 : DEC_GUIDE_MICROSTEPPING);
 }
     #endif
 
     #if RA_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART || DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART
+/**
+ * Runs motor at specified speed for specified duration in each direction, allowing 0.5s to stop motion after each move
+ * @param[in,out] driver The stepper driver instance
+ * @param[in] speed The speed to run the stepper motor at (in deg/s? idk)
```
julianneswinoga: Could you modify the comment to reflect what the function does? I just wrote out a basic structure that you could fill in

julianneswinoga (changes requested):

julianneswinoga (approved):

PR #84, opened 2021-03-27
Fix default TMC2209 serial port for Mega2560

julianneswinoga:
As per the wiring diagram on the wiki, the default serial port should be `Serial3` and not `Serial2`

Mega2560 Wiki Wiring

### DIFF
--- a/boards/AVR_MEGA2560/pins_MEGA2560.hpp
+++ b/boards/AVR_MEGA2560/pins_MEGA2560.hpp
@@ -53,7 +53,7 @@
 #endif    
 // DRIVER_TYPE_TMC2209_UART HardwareSerial port, can be shared across all drivers
 #ifndef RA_SERIAL_PORT
-  #define RA_SERIAL_PORT Serial2
+  #define RA_SERIAL_PORT Serial3
 #endif
 #ifndef RA_DRIVER_ADDRESS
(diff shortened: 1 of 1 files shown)


julianneswinoga:
@ClutchplateDude @andre-stefanov ?

andre-stefanov:
Mh the question here is what is better to change ... implementation or wiki. Changing implementation will introduce a breaking change to already existing builds ...

julianneswinoga:
Yeah I know... My reasoning for making a functional change and not a documentation one:

1. Users build according to the wiki first and then figure out the firmware
2. Until the UART test was added by @CTetford I had thought everything was working when it actually wasn't. Motors still were moving/tracking etc, there was no outward indication that it wasn't working
3. Current OAT builds that are in the same boat (UART isn't working, unbeknownst to the user) will now have a working UART connection

Of course I leave it up to those that have a longer history with the project, but the above reasoning makes sense to me

ClutchplateDude:
I think it's fine to change implementation instead of redoing the wiring diagrams.... I also had to add an override in my local config, but as I noticed yesterday that test does not work for me....

ClutchplateDude:
Hey Cameron, are you Cam S on Discord?

CTetford (approved):

PR #85, opened 2021-04-05
V1.9.03 - Updates

ClutchplateDude:
- Removed connection test check and made it standard.
- Switched motors to normal mode (better perf), but allowed silent mode through XX_UART_STEALTH_MODE defines.
- Made DEC driver be in Guide microstep mode unless slewing.
- Added command to Meade protocol to calculate stepper positions for stellar coordinates.
- Fixed an overflow bug in the CAL LCD menu for RA/DEC steps (esoteric mounts).

### DIFF
--- a/ConfigurationValidation.hpp
+++ b/ConfigurationValidation.hpp
@@ -48,9 +48,7 @@
     // Serial bus address must be specified for TMC2209 in UART mode
     #error RA driver address for DRIVER_TYPE_TMC2209_UART not specified.
   #endif
-  #ifndef RA_UART_STEALTH_MODE 
-    #error RA stealth mode not set! 
-  #elif (RA_UART_STEALTH_MODE != 0) && (RA_UART_STEALTH_MODE != 1)
+  #if (RA_UART_STEALTH_MODE != 0) && (RA_UART_STEALTH_MODE != 1)
     // Stealth mode must be zero or 1
     #error RA stealth mode must be 0 (off) or 1 (on)
   #endif
@@ -61,16 +59,14 @@
     // Serial bus address must be specified for TMC2209 in UART mode
     #error DEC driver address for DRIVER_TYPE_TMC2209_UART not specified.
   #endif
-  #ifndef DEC_UART_STEALTH_MODE 
-    #error DEC stealth mode not set! 
-  #elif (DEC_UART_STEALTH_MODE != 0) && (DEC_UART_STEALTH_MODE != 1)
+  #if (DEC_UART_STEALTH_MODE != 0) && (DEC_UART_STEALTH_MODE != 1)
     // Stealth mode must be zero or 1
     #error DEC stealth mode must be 0 (off) or 1 (on)
   #endif
 #endif
 
 #ifdef UART_CONNECTION_TEST_TXRX
-    #error The UART Connection test is now standard. This directive is obsolete and can be removed from your configuration.
+    #warning The UART Connection test is now standard. This directive is obsolete and can be removed from your configuration.
 #endif
 
 #if (AZIMUTH_ALTITUDE_MOTORS == 0)
--- a/ConfigurationValidation.hpp
+++ b/ConfigurationValidation.hpp
@@ -48,6 +48,12 @@
     // Serial bus address must be specified for TMC2209 in UART mode
     #error RA driver address for DRIVER_TYPE_TMC2209_UART not specified.
   #endif
+  #ifndef RA_UART_STEALTH_MODE 
+    #error RA stealth mode not set! 
+  #elif (RA_UART_STEALTH_MODE != 0) && (RA_UART_STEALTH_MODE != 1)
+    // Stealth mode must be zero or 1
+    #error RA stealth mode must be 0 (off) or 1 (on)
+  #endif
 #endif
 
 #if (DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
@@ -55,6 +61,16 @@
     // Serial bus address must be specified for TMC2209 in UART mode
     #error DEC driver address for DRIVER_TYPE_TMC2209_UART not specified.
   #endif
+  #ifndef DEC_UART_STEALTH_MODE 
+    #error DEC stealth mode not set! 
+  #elif (DEC_UART_STEALTH_MODE != 0) && (DEC_UART_STEALTH_MODE != 1)
+    // Stealth mode must be zero or 1
+    #error DEC stealth mode must be 0 (off) or 1 (on)
+  #endif
+#endif
+
+#ifdef UART_CONNECTION_TEST_TXRX
+    #error The UART Connection test is now standard. This directive is obsolete and can be removed from your configuration.
 #endif
 
 #if (AZIMUTH_ALTITUDE_MOTORS == 0)
--- a/ConfigurationValidation.hpp
+++ b/ConfigurationValidation.hpp
@@ -65,10 +65,6 @@
   #endif
 #endif
 
-#ifdef UART_CONNECTION_TEST_TXRX
-    #warning The UART Connection test is now standard. This directive is obsolete and can be removed from your configuration.
-#endif
-
 #if (AZIMUTH_ALTITUDE_MOTORS == 0)
   // Baseline configuration without azimuth & altitude control is valid
 #elif defined(__AVR_ATmega2560__)
--- a/Configuration_adv.hpp
+++ b/Configuration_adv.hpp
@@ -156,6 +156,12 @@
     #define USE_VREF 0      //By default Vref is ignored when using UART to specify rms current. Only enable if you know what you are doing.
   #endif
   
+  #ifndef UART_CONNECTION_TEST_TXRX
+    #define UART_CONNECTION_TEST_TXRX 0
+  #endif
+  #ifndef UART_CONNECTION_TEST_TX
+    #define UART_CONNECTION_TEST_TX 0
+  #endif  
 #endif
 
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/Configuration_adv.hpp
+++ b/Configuration_adv.hpp
@@ -97,6 +97,9 @@
   #ifndef RA_TRACKING_MICROSTEPPING 
     #define RA_TRACKING_MICROSTEPPING 8     // The fine microstep mode for tracking RA axis
   #endif
+  #ifndef RA_UART_STEALTH_MODE
+    #define RA_UART_STEALTH_MODE   0
+  #endif
 #elif (RA_DRIVER_TYPE == DRIVER_TYPE_A4988_GENERIC) || (RA_DRIVER_TYPE == DRIVER_TYPE_TMC2209_STANDALONE)
   #define RA_SLEW_MICROSTEPPING 8         // Microstep mode set by MS pin strapping. Use the same microstep mode for both slewing & tracking   
   #define RA_TRACKING_MICROSTEPPING RA_SLEW_MICROSTEPPING   
@@ -113,6 +116,9 @@
   #endif
   #ifndef DEC_GUIDE_MICROSTEPPING 
     #define DEC_GUIDE_MICROSTEPPING 16  // The fine microstep mode used for guiding DEC only
+  #endif
+  #ifndef DEC_UART_STEALTH_MODE
+    #define DEC_UART_STEALTH_MODE   0
   #endif
 #elif (DEC_DRIVER_TYPE == DRIVER_TYPE_A4988_GENERIC) || (DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_STANDALONE)
   #define DEC_SLEW_MICROSTEPPING  16  // Only UART drivers support dynamic switching. Use the same microstep mode for both slewing & guiding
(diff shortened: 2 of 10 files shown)


Configuration_adv.hpp
```diff
@@ -150,12 +156,6 @@
     #define USE_VREF 0      //By default Vref is ignored when using UART to specify rms current. Only enable if you know what you are doing.
   #endif
   
-  #ifndef UART_CONNECTION_TEST_TXRX
```
CTetford: These lines shouldn't be removed - they're for checking bidirectional UART connection on boards with TX and RX wired.

boards/AVR_MKS_GEN_L_V21/pins_MKS_GEN_L_V21.h
```diff
@@ -95,7 +95,6 @@
 #endif
 
 #define SW_SERIAL_UART 1
-#define UART_CONNECTION_TEST_TXRX 1
```
CTetford: Same as above

src/Mount.hpp
```diff
@@ -328,15 +330,17 @@ class Mount {
   DayTime calculateLst();
   DayTime calculateHa();
   
-  #if UART_CONNECTION_TEST_TX == 1
```
CTetford, line 331: The UART_CONNECTION_TEST_TX flags shouldn't be removed - this should still be a configurable item.

CTetford: Furthermore, for boards with both TX/RX like the MKS v2.x, the UART_CONNECTION_TEST_TX routine should be superseded by the UART_CONNECTION_TEST_TXRX routine. Hence why it needs to be configurable

ConfigurationValidation.hpp
```diff
@@ -48,13 +48,29 @@
...
+    #error DEC stealth mode not set! 
+  #elif (DEC_UART_STEALTH_MODE != 0) && (DEC_UART_STEALTH_MODE != 1)
+    // Stealth mode must be zero or 1
+    #error DEC stealth mode must be 0 (off) or 1 (on)
+  #endif
+#endif
+
+#ifdef UART_CONNECTION_TEST_TXRX
+    #error The UART Connection test is now standard. This directive is obsolete and can be removed from your configuration.
 #endif
```
andre-stefanov: i would not throw an error if there is an unused definition in the config. this is not breaking anything. at some point people will generate a new config anyway.

ConfigurationValidation.hpp
```diff
@@ -48,13 +48,29 @@
...
+  #endif
 #endif
 
 #if (DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
   #ifndef DEC_DRIVER_ADDRESS
     // Serial bus address must be specified for TMC2209 in UART mode
     #error DEC driver address for DRIVER_TYPE_TMC2209_UART not specified.
   #endif
+  #ifndef DEC_UART_STEALTH_MODE 
+    #error DEC stealth mode not set!
```
andre-stefanov: can this ever happen? we are setting this definition in Configuration_adv in case its missing

src/Mount.cpp
```diff
@@ -335,12 +335,11 @@ void Mount::configureDECStepper(byte pin1, byte pin2, int maxSpeed, int maxAccel
...
 #endif
 
-#if RA_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART || DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART
-#if UART_CONNECTION_TEST_TXRX == 1
+#if RA_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART || DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART || AZ_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART || ALT_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART 
 bool Mount::connectToDriver( TMC2209Stepper* driver, const char *driverKind ) {
     LOGV2(DEBUG_STEPPERS, F("Testing UART Connection to %s driver..."), driverKind);
-    for(int i=0; i<5; i++) {
-        if(driver->test_connection() == 0) {
+    for(int i = 0; i < 5; i++) {
```
andre-stefanov: mh ... we should give this "magic" 5 a name somewhere. e.g. UART_CONNECTION_TEST_RETRIES

src/Mount.cpp
```diff
@@ -335,12 +335,11 @@ void Mount::configureDECStepper(byte pin1, byte pin2, int maxSpeed, int maxAccel
...
+        if (driver->test_connection() == 0) {
```
andre-stefanov, line 345: the return value of test_connection varies in different situations (indicating a bit more details on the actual error). we should use these and display to the user the possible reason. [Here](https://github.com/teemuatlut/TMCStepper/blob/ddc0c121af9117c14c173427f1dae165d8409ea5/examples/Calibrate_spreadCycle/Calibrate_spreadCycle.ino#L94) is an example.

src/Mount.cpp
```diff
@@ -335,12 +335,11 @@ void Mount::configureDECStepper(byte pin1, byte pin2, int maxSpeed, int maxAccel
   #endif  
 #endif
 
-#if RA_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART || DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART
-#if UART_CONNECTION_TEST_TXRX == 1
+#if RA_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART || DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART || AZ_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART || ALT_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART 
 bool Mount::connectToDriver( TMC2209Stepper* driver, const char *driverKind ) {
```
andre-stefanov, line 342: `driverKind` is only used for logging. it can be also printed by the caller

src/Mount.cpp
andre-stefanov: In near future displaying this on the display would be a nice improvement (since people dont debug the oat while in the field)

src/Mount.cpp
```diff
@@ -365,25 +363,22 @@ bool Mount::connectToDriver( TMC2209Stepper* driver, const char *driverKind ) {
...
+        digitalWrite(RA_EN_PIN, HIGH);    // Disable motor for safety reasons if UART connection fails to avoid operating at incorrect rms_current
+    }
     _driverRA->toff(0);    
     #if USE_VREF == 0  //By default, Vref is ignored when using UART to specify rms current.
       _driverRA->I_scale_analog(0);
     #endif
     LOGV2(DEBUG_STEPPERS, F("Mount: Requested RA motor rms_current: %d mA"), rmscurrent);
     _driverRA->rms_current(rmscurrent, 1.0f); //holdMultiplier = 1 to set ihold = irun
     _driverRA->toff(1);
+    _driverRA->en_spreadCycle(1 - RA_UART_STEALTH_MODE);
```
andre-stefanov: why `1 - value`? `en_spreadCycle` expects a boolean as parameter, meaning we should rather use `!RA_UART_STEALTH_MODE` or `RA_UART_STEALTH_MODE == 0`

src/Mount.cpp
```diff
@@ -394,33 +389,33 @@ bool Mount::connectToDriver( TMC2209Stepper* driver, const char *driverKind ) {
...
+    }
+
     _driverRA->toff(0);   
     #if USE_VREF == 0  //By default, Vref is ignored when using UART to specify rms current.
         _driverRA->I_scale_analog(0);
     #endif
     LOGV2(DEBUG_STEPPERS, F("Mount: Requested RA motor rms_current: %d mA"), rmscurrent);
     _driverRA->rms_current(rmscurrent, 1.0f); //holdMultiplier = 1 to set ihold = irun
     _driverRA->toff(1);
+    _driverRA->en_spreadCycle(1 - RA_UART_STEALTH_MODE);
```
andre-stefanov: same as above

src/Mount.cpp
```diff
@@ -2115,6 +2087,13 @@ void Mount::loop() {
             startSlewing(TRACKING);					   
           }
         #endif
+
+        // Reset DEC to guide microstepping so that guiding is always ready and no switch is neccessary on guide pulses.
+        #if DEC_STEPPER_TYPE == STEPPER_TYPE_NEMA17 && DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART
```
andre-stefanov: do we really need to check for the stepper type here? TMC2209 is being used with nemas only

andre-stefanov: and even if its something else ... the actual logic is based on the driver only

src/Mount.cpp
```diff
@@ -2332,6 +2311,21 @@ float Mount::getSpeed(int direction) {
   return 0;
 }
 
+/////////////////////////////////
+//
+// calculatePositions
+//
+// This code calculates the stepper locations to move to, given the right ascension and declination
+/////////////////////////////////
+void Mount::calculatePositions(float raCoord, float decCoord, long& raPos, long& decPos){
```
andre-stefanov: I would rather name it `calculateTargetPositions` and the param names accordingly. Otherwise it can be misunderstood as "current position"

src/c76_menuCAL.hpp
```diff
@@ -168,6 +168,17 @@ bool checkProgressiveUpDown(int *val, int minDelay = 25)
   return ret;
 }
 
+bool checkProgressiveUpDown(int *val, int minDelay = 25)
+{
+  long lval = *val;
+  bool ret = checkProgressiveUpDown(&lval, minDelay);
+  if ((lval < 32766L) && (lval > -32765L))
```
andre-stefanov: please use `INT16_MIN` and `INT16_MAX` instead of magic numbers

src/c76_menuCAL.hpp
```diff
@@ -168,6 +168,17 @@ bool checkProgressiveUpDown(int *val, int minDelay = 25)
   return ret;
 }
 
+bool checkProgressiveUpDown(int *val, int minDelay = 25)
```
andre-stefanov, line 171: i dont understand this overload. we allow passing an int while handling it as a long. why is it needed? also this could lead to weird/invalid memory access. Especially since the value behind it is being modified by a side effect.

andre-stefanov (changes requested):

CTetford (changes requested):

ConfigurationValidation.hpp
ClutchplateDude: Fair enough. I'll make it a warning

ConfigurationValidation.hpp
ClutchplateDude: Not really...

src/Mount.cpp
ClutchplateDude: Oh, all the code I saw used 0 or 1, lol

src/Mount.cpp
ClutchplateDude: True

src/Mount.cpp
ClutchplateDude: Well.... it doesn't calculate the target position of the mount, just as it doesn't calculate the current position. How about calculateStepperPositions?

src/c76_menuCAL.hpp
ClutchplateDude: I could not get any constants to compile across all platforms, which is why I ended up using these...

src/c76_menuCAL.hpp
ClutchplateDude: There are many functions that use this to handle ints, and one that handles long. I don't understand how this can possibly lead to weird/invalid memory access. This simply allows longs and ints to use this function and we stop an int from overflowing.

src/Mount.cpp
ClutchplateDude: HAL can do that....

src/Mount.cpp
ClutchplateDude: Fixed

src/Mount.cpp
ClutchplateDude: Fixed

src/Mount.cpp
ClutchplateDude: Fixed

src/Mount.cpp
ClutchplateDude: Fixed

src/c76_menuCAL.hpp
ClutchplateDude: Fixed

ConfigurationValidation.hpp
ClutchplateDude: Removed defensive code.

ConfigurationValidation.hpp
ClutchplateDude: Fixed

src/Mount.cpp
ClutchplateDude: Won't Fix

src/c76_menuCAL.hpp
andre-stefanov: convinced after discussion on discord.

andre-stefanov (approved):

ConfigurationValidation.hpp
julianneswinoga: @ClutchplateDude FYI `#warning`s still cause a build failure, you should use `#pragma message "stuff"`

CTetford (approved):

PR #91, opened 2021-04-19
Added new DISPLAY_TYPE 

spuduk:
Added a new DISPLAY_TYPE

DISPLAY_TYPE_LCD_KEYPAD_I2C_SSD1306

This edit is to enable use of the SSD1306 128x32 OLED display and a keypad instead of the joystick on the Mega2560. Still waiting on HW but tested working with original keypad via jumpers.

Keypad: https://www.banggood.com/AD-Analog-Keyboard-Module-Electronic-Building-Blocks-5-Keys-DIY-p-1374279.html
Display: https://www.banggood.com/Geekcreit-0_91-Inch-128x32-IIC-I2C-Blue-OLED-LCD-Display-DIY-Module-SSD1306-Driver-IC-DC-3_3V-5V-p-1140506.html


julianneswinoga:
It looks like there are no functional changes in order to support the keypad as the hardware is identical to the joystick. Is that correct?

spuduk:
Yes, The changes utilise a combination of the existing oled display and lcd keypad code. (I2C Display and resistor ladder switches connected to A0)
.
(The joystick requires 3 analog pins, the keypad requires 1)

julianneswinoga:
I'm still a little confused by the changes, do you not get identical functionality by just setting `#define DISPLAY_TYPE DISPLAY_TYPE_LCD_JOY_I2C_SSD1306`?

spuduk:
#define DISPLAY_TYPE DISPLAY_TYPE_LCD_JOY_I2C_SSD1306 wont compile on a MEGA2560 due to the pin assignments missing for the JOYSTICK. 

You can add the pin assignments for the JOYSTICK but for the KEYPAD you do not require those pins.

julianneswinoga:
Ok, I think the cleanest way that we can support this functionality would be to break up the keypad and joystick definitions. That would avoid creating a new hardware definition that doesn't add any functionality. In it's current state, I don't like the changes. Would you be comfortable splitting up that definition @spuduk ?

ClutchplateDude:
Looks like this is abandoned.

PR #92, opened 2021-04-20
V1.9.07 - Updates

ClutchplateDude:
- Split AZ and ALT motors, so you can have one but not the other.

Not sure I got matrix.py correct....

Will still need to update OATControl.


christian-kardach:
Couple of findings as I tested with TMC2209 and a NEMA14, both Alt and AZ
Spreadcyckle is on, not sure its needed for Az/Alt?
MiniControl button release doesnt seem to send correct command/or doesnt work in OC/FW, it just keeps spinning for 30 seconds

ClutchplateDude:
> Couple of findings as I tested with TMC2209 and a NEMA14, both Alt and AZ
> Spreadcyckle is on, not sure its needed for Az/Alt?
> MiniControl button release doesnt seem to send correct command/or doesnt work in OC/FW, it just keeps spinning for 30 seconds

Yeah, sorry, should have explained... it runs a certain number of degrees, as determined by the slew rate. So press '1', '2', '3' or '4' for different distances. 4 for very rough and then 2 and 1 for fine adjustments.
I only have the BYJ48 for testing, and the NEMA stuff was Camerons initial code, so no change. Spreadcycle must have been set before as well. right?

andre-stefanov (approved):

PR #94, opened 2021-04-28
AutoPA should be deactivated by default

christian-kardach:


ClutchplateDude:
No-one should be using the sample anymore.... maybe we should remove it..... but this is a good for now.

ClutchplateDude (approved):

Changelog.md
```diff
@@ -1,3 +1,6 @@
+**V1.9.09 - Updates**
+- Sample Local Config - Disabled as default
```
ClutchplateDude, line 2: "Disabled AZ/ALT steppers by default."

PR #95, opened 2021-05-01
V1.9.10 - Updates

ClutchplateDude:
- Fixed a compiler bug when enabling LCD_BUTTON_TEST
- Allowed connection test to be set in local config

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,5 @@
 **V1.9.10 - Updates**
-- Fixed a compiler bug when enabling LCD_BUTTON_TEST
+- Fixed a bug that caused a compilation fail when enabling LCD_BUTTON_TEST
 - Allowed connection test to be set in local config
 
 **V1.9.09 - Updates**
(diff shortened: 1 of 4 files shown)


src/c_buttons.hpp
```diff
@@ -21,7 +22,7 @@
   int lastLoopKey = -1;
 
   #if LCD_BUTTON_TEST == 1
```
julianneswinoga, line 24: We should add the button test to the matrix build to ensure this won't break in the future

Changelog.md
```diff
@@ -1,5 +1,9 @@
+**V1.9.10 - Updates**
+- Fixed a compiler bug when enabling LCD_BUTTON_TEST
```
julianneswinoga: I would also say "compilation issue" rather than "compiler bug"... Thought you found an issue with GCC at first read :laughing:

ClutchplateDude: LOL, totally.. must have been a lack of coffee...

src/c_buttons.hpp
ClutchplateDude: Yeah, agreed. @andre-stefanov would this be easy to do?

CTetford (approved):

PR #96, opened 2021-05-03
Fix Gyro hangs

fzsombor:
The I2C communication in the Wire.h library is implemented with while loops. If the physical wiring of the I2C component introduces stale readings from the pins the functions could stick in the while loop so the function never returns. This would cause the hang of the whole FW of the OAT.

Changes:
* Implemented the timeout feature in the FW so it cannot hang anymore
* Set the clock rate of the I2C communication to the lowest setting to increase stability with imperfect wiring (too long cables, no pull-up resistors, etc.)
* If a timeout happens it is now logged on the INFO level
* Fixed some typos

### DIFF
--- a/src/Gyro.cpp
+++ b/src/Gyro.cpp
@@ -27,6 +27,8 @@ void Gyro::startup()
     // Initialize interface to the MPU6050
     LOGV1(DEBUG_INFO, F("GYRO:: Starting"));
     Wire.begin();
+    Wire.setClock(100000);          // Set lowest clock speed to make the communication more reliable
+    Wire.setWireTimeout(3000, true); // Set I2C timeout if the Wire.h calls are hanging due to a HW issue 
 
     // Execute 1 byte read from MPU6050_REG_WHO_AM_I
     // This is a read-only register which should have the value 0x68
@@ -34,7 +36,7 @@ void Gyro::startup()
     Wire.write(MPU6050_REG_WHO_AM_I);
     Wire.endTransmission(true);
     Wire.requestFrom(MPU6050_I2C_ADDR, 1, 1);
-    byte id = (Wire.read() >> 1) & 0x3F;    
+    byte id = (Wire.read() >> 1) & 0x3F;
     isPresent = (id == 0x34);
     if (!isPresent) {
         LOGV1(DEBUG_INFO, F("GYRO:: Not found!"));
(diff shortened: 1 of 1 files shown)


ClutchplateDude (approved):

ClutchplateDude:
This is cool, thanks for looking at this!

ClutchplateDude:
This does not compile for Mega2560... @fzsombor can you check into this?

fzsombor:
Is there any output from PlatformIO? It is compiling in Arduino IDE, and I tested it on my Mega but I don’t have a Windows machine close.

fzsombor:
It compiles for every other board because gyro is only enabled for Mega. I think this is a mismatch between the Wire.h library.

fzsombor:
Yeah, MegaCore does not contain the changes made in the upstream ArduinoCore
https://github.com/MCUdude/MegaCore/blob/master/avr/libraries/Wire/src/Wire.h
vs.
https://github.com/arduino/ArduinoCore-avr/blob/master/libraries/Wire/src/Wire.h

Not sure what to do with this as Arduino IDE compiles it.

src/Gyro.cpp
```diff
@@ -27,14 +27,16 @@ void Gyro::startup()
     // Initialize interface to the MPU6050
     LOGV1(DEBUG_INFO, F("GYRO:: Starting"));
     Wire.begin();
+    Wire.setClock(100000);          // Set lowest clock speed to make the communication more reliable
+    Wire.setWireTimeout(3000, true); // Set I2C timeout if the Wire.h calls are hanging due to a HW issue
```
ClutchplateDude, line 31: I think you must be using a different library. These functions error out on my MKS build

ClutchplateDude: Ok, it pays to read previous conversations. I have Gyro enabled on my MKS, that's probably why it fails? I guess we should switch to the newer library on PIO.

fzsombor: @ClutchplateDude if you are trying to use this in the software I2C with MKS, that's not gonna work. You are using a library which is a wrapper around the soft I2C library and acts as an interface for the functions of Wire.h. In this wrapper these new functions are not defined either, because that is based also on the old Wire.h. (See https://github.com/felias-fogg/SlowSoftWire/blob/master/SlowSoftWire.h vs. https://github.com/arduino/ArduinoCore-avr/blob/master/libraries/Wire/src/Wire.h)

ClutchplateDude:
We're going to skip this for now, due to library incompatibilities.

PR #97, opened 2021-05-04
Support for Focus motor on MKS boards

ClutchplateDude:
Added support for a Focuser stepper that supports the Meade LX200 command set (:Fxxx family).


ConfigurationValidation.hpp
```diff
@@ -145,6 +145,44 @@
...
+
+  #if (FOCUS_STEPPER_TYPE == DRIVER_TYPE_TMC2209_UART)
+    #ifndef FOCUS_DRIVER_ADDRESS
+      // Serial bus address must be specified for TMC2209 in UART mode
+      #error Focus driver address for DRIVER_TYPE_TMC2209_UART not specified.
+    #endif
+  #endif
+
+#else
+  #error Configuration does not support Focus. Use at own risk.
```
julianneswinoga: Warning instead of error

ConfigurationValidation.hpp
```diff
@@ -145,6 +145,44 @@
...
+  #elif (FOCUS_STEPPER_TYPE == STEPPER_TYPE_28BYJ48) && (FOCUS_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
+    // Valid ALT stepper and driver combination
+  #elif (FOCUS_STEPPER_TYPE == STEPPER_TYPE_NEMA17) && (FOCUS_DRIVER_TYPE == DRIVER_TYPE_A4988_GENERIC)
+    // Valid ALT stepper and driver combination
+  #elif (FOCUS_STEPPER_TYPE == STEPPER_TYPE_NEMA17) && (FOCUS_DRIVER_TYPE == DRIVER_TYPE_TMC2209_STANDALONE)
+    // Valid ALT stepper and driver combination
+  #elif (FOCUS_STEPPER_TYPE == STEPPER_TYPE_NEMA17) && (FOCUS_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
+    // Valid Focus stepper and driver combination
+  #else
+    #error Unsupported Focus stepper & driver combination. Use at own risk.
```
julianneswinoga, line 172: Warning instead of error

Configuration_sample_local.hpp
```diff
@@ -86,6 +86,23 @@
...
 #define ALT_OPERATING_CURRENT_SETTING 100  // ALT operating setting as a percentage of motor rating
 
+/**
+ * @brief Focus
+ * Set FOCUS_STEPPER_TYPE to the correct stepper type to 
+ * enable, or to STEPPER_TYPE_NONE to exclude Focus from configuration.
+ */
+#define FOCUS_STEPPER_TYPE    STEPPER_TYPE_NEMA17
+#define FOCUS_DRIVER_TYPE     DRIVER_TYPE_TMC2209_UART
+#define FOCUS_CORRECTION_FACTOR 1.000f
```
julianneswinoga: Is this not used anywhere?

boards/AVR_MKS_GEN_L_V21/pins_MKS_GEN_L_V21.h
```diff
@@ -95,7 +95,7 @@
 #endif
 
 #define SW_SERIAL_UART 1
-#ifndef UART_CONNECTION_TEST_TXRX
+#ifndef UART_CONNECTION_TEST_TXRX
```
julianneswinoga: You should configure your editor to strip trailing whitespace so we don't get random changes like this

Configuration_sample_local.hpp
```diff
@@ -86,6 +86,23 @@
...
+
+// TMC2209 UART settings
+// These settings work only with TMC2209 in UART connection (single wire to TX)
+#define FOCUS_MOTOR_CURRENT_RATING         0     // Current rating of focus motor in mA
+#define FOCUS_OPERATING_CURRENT_SETTING  100     // Operating setting as a percentage of focus motor rating
+#define FOCUS_STEPPER_SPEED              200     // Default speed when moving focus motor in steps/s
+#define FOCUS_DISABLE_TIME                 5     // Disable Focus motor after X seconds of idle
```
julianneswinoga: Not used either

src/Mount.cpp
```diff
@@ -45,6 +45,11 @@ POP_NO_WARNINGS
 #define SLEW_MASK_WEST    B1000
 #define SLEW_MASK_ANY     B1111
 
+// Focuser support
+#define FOCUS_IDLE       0
```
julianneswinoga: Make these an enum please

src/Mount.cpp
```diff
@@ -664,6 +702,81 @@ bool Mount::connectToDriver( TMC2209Stepper* driver, const char *driverKind ) {
...
+// TMC2209 UART only
+/////////////////////////////////
+#if (FOCUS_STEPPER_TYPE != STEPPER_TYPE_NONE) && (FOCUS_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
+#if SW_SERIAL_UART == 0
+  void Mount::configureFocusDriver(Stream *serial, float rsense, byte driveraddress, int rmscurrent, int stallvalue)
+  {
+    _driverFocus = new TMC2209Stepper(serial, rsense, driveraddress);
+    _driverFocus->begin();
+    bool _UART_Rx_connected = false;
+    #if UART_CONNECTION_TEST_TXRX == 1
```
julianneswinoga, line 713: Can you refactor this part to something like:
```C++
    #if UART_CONNECTION_TEST_TXRX == 1
        const bool UART_Rx_connected = connectToDriver( _driverFocus, "Focus" );
        if (!UART_Rx_connected) {
            digitalWrite(FOCUS_EN_PIN, HIGH);    //Disable motor for safety reasons if UART connection fails to avoid operating at incorrect rms_current
        }
    if (UART_Rx_connected){
        LOGV2(DEBUG_STEPPERS, F("Mount: Actual Focus motor rms_current: %d mA"), _driverFocus->rms_current());
        LOGV2(DEBUG_STEPPERS, F("Mount: Actual Focus CS value: %d"), _driverFocus->cs_actual());
        LOGV2(DEBUG_STEPPERS, F("Mount: Actual Focus vsense: %d"), _driverFocus->vsense());
    }
    #endif
```

- Remove leading underscore for local variable (it's a reserved identifier)
- Move local var into `#if` block since that's the only place where it's used

src/Mount.cpp
```diff
@@ -664,6 +702,81 @@ bool Mount::connectToDriver( TMC2209Stepper* driver, const char *driverKind ) {
...
 #endif
 
+/////////////////////////////////
+//
+// configureFocusdriver
+// TMC2209 UART only
+/////////////////////////////////
+#if (FOCUS_STEPPER_TYPE != STEPPER_TYPE_NONE) && (FOCUS_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
+#if SW_SERIAL_UART == 0
+  void Mount::configureFocusDriver(Stream *serial, float rsense, byte driveraddress, int rmscurrent, int stallvalue)
```
julianneswinoga, line 709: You should refactor both instances of the serial into one, there's only a minor difference to the serial initialization

src/Mount.cpp
```diff
@@ -664,6 +702,81 @@ bool Mount::connectToDriver( TMC2209Stepper* driver, const char *driverKind ) {
...
+        if (!_UART_Rx_connected) {
+            digitalWrite(ALT_EN_PIN, HIGH);    //Disable motor for safety reasons if UART connection fails to avoid operating at incorrect rms_current
+        }
+    #endif
+    _driverFocus->toff(0);
+    #if USE_VREF == 0  //By default, Vref is ignored when using UART to specify rms current.
+        _driverFocus->I_scale_analog(0);
+    #endif
+    LOGV2(DEBUG_STEPPERS, F("Mount: Requested Focus motor rms_current: %d mA"), rmscurrent);
+    _driverFocus->rms_current(rmscurrent, 0.1f); //holdMultiplier = 1 to set ihold = irun
```
julianneswinoga, line 725: Is the `0.1f` meant to be different to the below `0.01f?

src/Mount.cpp
```diff
@@ -1447,6 +1566,52 @@ void Mount::setSpeed(int which, float speedDegsPerSec) {
     #endif
   }
   #endif
+  
+  #if (FOCUS_STEPPER_TYPE != STEPPER_TYPE_NONE)
+    else if (which == FOCUS_STEPS) {
```
julianneswinoga: This indentation doesn't match the previous, dedent by 2 spaces

src/Mount.cpp
```diff
@@ -1447,6 +1566,52 @@ void Mount::setSpeed(int which, float speedDegsPerSec) {
...
+          waitUntilStopped(FOCUSING);
+        }
+
+        // Are we starting a move or changing speeds?
+        if (speedDegsPerSec != 0) 
+        {
+          enableFocusMotor();
+          _stepperFocus->setMaxSpeed(speedDegsPerSec);
+          _stepperFocus->moveTo(sign(speedDegsPerSec) * 300000);
+          _focuserMode = FOCUS_TO_TARGET; //?
```
julianneswinoga: `//?` ?

src/Mount.cpp
```diff
@@ -1603,6 +1768,126 @@ void Mount::enableAzAltMotors() {
...
+      //_stepperFocus->setSpeed(speedFactor[_focusRate ] * _maxFocusSpeed);
+    }
+}
+
+/////////////////////////////////
+//
+// focusContinuousMove
+//
+/////////////////////////////////
+void Mount::focusContinuousMove(int direction)
```
julianneswinoga: We should definitely have a direction enum, currently you could call this with any number which would change the applied speed

src/Mount.cpp
```diff
@@ -1689,7 +1974,7 @@ String Mount::getStatusString() {
     status = "Idle,";
   }
 
-  String disp = "-----,";
+  String disp = "------,";
```
julianneswinoga, line 1977: Won't this break the `:GX` meade command?

src/Mount.cpp
```diff
@@ -2087,6 +2390,26 @@ void Mount::loop() {
   }
   #endif
 
+  #if (FOCUS_STEPPER_TYPE != STEPPER_TYPE_NONE)
+    // LOGV2(DEBUG_MOUNT, F("Mount: Focuser running:  %d"), _stepperFocus->isRunning());
+
+    if(_stepperFocus->isRunning())
+    {
+      LOGV2(DEBUG_MOUNT, F("Mount: Focuser running at speed %f"), _stepperFocus->speed());
+      _lastFocusMovementTimestamp = now;
```
julianneswinoga: `_lastFocusMovementTimestamp` isn't read from anywhere

src/Mount.cpp
```diff
@@ -2087,6 +2390,26 @@ void Mount::loop() {
...
+    }
+    else if (_focuserWasRunning)
```
julianneswinoga, line 2401: `_focuserWasRunning` is always `false`, this code will never run

src/c77_menuFOC.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+#if (DISPLAY_TYPE > 0) && (FOCUS_STEPPER_TYPE != STEPPER_TYPE_NONE)
+
+// HIGHLIGHT states allow you to pick one of the sub functions.
+enum
+{
+  HIGHLIGHT_FOCUS_FIRST = 1,
+  HIGHLIGHT_FOCUS_ADJUSTMENT = 1,
+
+  HIGHLIGHT_FOCUS_LAST = HIGHLIGHT_FOCUS_ADJUSTMENT
```
julianneswinoga: Unused

src/c77_menuFOC.hpp
```diff
@@ -0,0 +1,120 @@
+#pragma once
+
+#if (DISPLAY_TYPE > 0) && (FOCUS_STEPPER_TYPE != STEPPER_TYPE_NONE)
+
+// HIGHLIGHT states allow you to pick one of the sub functions.
+enum
```
julianneswinoga: Named enum

src/c77_menuFOC.hpp
```diff
@@ -0,0 +1,120 @@
...
+{
+  HIGHLIGHT_FOCUS_FIRST = 1,
+  HIGHLIGHT_FOCUS_ADJUSTMENT = 1,
+
+  HIGHLIGHT_FOCUS_LAST = HIGHLIGHT_FOCUS_ADJUSTMENT
+};
+
+#define FOCUS_ADJUSTMENT 90
```
julianneswinoga: Why make an enum and then just not use it?

src/c77_menuFOC.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+byte focState = HIGHLIGHT_FOCUS_FIRST;
```
julianneswinoga: Don't use raw numbers for state machines please! :)

src/c77_menuFOC.hpp
```diff
@@ -0,0 +1,120 @@
...
+{
+  char scratchBuffer[20];
+  if (focState == HIGHLIGHT_FOCUS_ADJUSTMENT)
+  {
+    lcdMenu.printMenu(">Focus Adjust");
+  }
+  else if (focState == FOCUS_ADJUSTMENT)
+  {
+    strcpy(scratchBuffer, "Rate:  1 2 3 4 *");
+    scratchBuffer[6 + rateIndex * 2] = '>';
```
julianneswinoga, line 105: There's got to be a better way to do this... This looks so brittle. Maybe split up the `Rate:` string from the options or something?

julianneswinoga (changes requested):

Configuration_sample_local.hpp
ClutchplateDude: Yeah, it was left over from earlier iterations.... Removed

ConfigurationValidation.hpp
ClutchplateDude: Done.

Configuration_sample_local.hpp
ClutchplateDude: Removed.

src/Mount.cpp

src/Mount.cpp
ClutchplateDude: Can't do that since the driver initialization needs to run before you can output the 3 values.

src/Mount.cpp
ClutchplateDude: Not in this PR. This pattern has been used for all the motors. We can revisit later, or once HAL is here.

src/Mount.cpp
ClutchplateDude: No, good catch!

src/Mount.cpp

src/Mount.cpp
ClutchplateDude: Removed

src/Mount.cpp

src/Mount.cpp
ClutchplateDude: No. It's not an official LX200 command and only OATControl uses it, and it doesn't care if there are extras.

src/Mount.cpp
ClutchplateDude: Yeah, left over from previous functionality.
Removed.

src/Mount.cpp
ClutchplateDude: Oops, good catch, fixed.

src/c77_menuFOC.hpp
ClutchplateDude: Yeah, it's just a pattern when there's  more than one item... Leaving for now.

src/c77_menuFOC.hpp
ClutchplateDude: ....

src/c77_menuFOC.hpp
ClutchplateDude: Fixed

src/c77_menuFOC.hpp
ClutchplateDude: We're trying to minimize memory allocations like ones that would result from string concatenations (prevent heap fragmentation). This is the best way.

src/Mount.cpp
```diff
@@ -1812,10 +1812,10 @@ void Mount::focusSetSpeedByRate(int rate)
 // focusContinuousMove
 //
 /////////////////////////////////
-void Mount::focusContinuousMove(int direction)
+void Mount::focusContinuousMove(FocuserDirection direction)
 {
   // maxSpeed is set to what the rate dictates
-  setSpeed(FOCUS_STEPS, direction * _maxFocusRateSpeed);
+  setSpeed(FOCUS_STEPS, (int)direction * _maxFocusRateSpeed);
```
julianneswinoga: You should use `static_cast<int>(` for compile-time type checking instead of C-style casts

src/c77_menuFOC.hpp
julianneswinoga: Yeah I know memory fragmentation is a bitch on constrained systems like this. Maybe I'll figure out a clean & performant way we can do string manipulation within menu option selections.

src/Mount.cpp
julianneswinoga: Fair, but duplication leads to copy+paste errors and missed patches. I always try to refactor in place where I'm already making changes

src/Mount.cpp
```diff
@@ -748,31 +747,33 @@ bool Mount::connectToDriver( TMC2209Stepper* driver, const char *driverKind ) {
...
-        }
+      bool UART_Rx_connected = false;
+      UART_Rx_connected = connectToDriver( _driverFocus, "Focus" );
+      if (!UART_Rx_connected) {
+          digitalWrite(FOCUS_EN_PIN, HIGH);    //Disable motor for safety reasons if UART connection fails to avoid operating at incorrect rms_current
+      }
     #endif
     _driverFocus->toff(0);
     #if USE_VREF == 0  //By default, Vref is ignored when using UART to specify rms current.
         _driverFocus->I_scale_analog(0);
```
julianneswinoga: Don't use integer literals for boolean values

src/Mount.cpp
```diff
@@ -748,31 +747,33 @@ bool Mount::connectToDriver( TMC2209Stepper* driver, const char *driverKind ) {
...
     #endif
     LOGV2(DEBUG_STEPPERS, F("Mount: Requested Focus motor rms_current: %d mA"), rmscurrent);
-    _driverFocus->rms_current(rmscurrent, 0.01f); //holdMultiplier = 1 to set ihold = irun
+    _driverFocus->rms_current(rmscurrent, 0.1f); //holdMultiplier = 1 to set ihold = irun
     _driverFocus->toff(1);
     _driverFocus->en_spreadCycle(FOCUS_UART_STEALTH_MODE == 0);
     _driverFocus->blank_time(24);
     _driverFocus->microsteps(FOCUS_MICROSTEPPING == 1 ? 0 : FOCUS_MICROSTEPPING);   // If 1 then disable microstepping
     _driverFocus->TCOOLTHRS(0xFFFFF);  //xFFFFF);
```
julianneswinoga, line 767: `//xFFFFF);` ?

julianneswinoga (changes requested):
Can you also make sure to squash before merging? There's a lot of commits for a (relatively) small changeset

src/Mount.cpp
ClutchplateDude: No idea. The hardware people wrote that :-)

src/Mount.cpp
ClutchplateDude: Making me fix legacy code.... tsk tsk

julianneswinoga: If no-one fixes it it'll stay legacy code :stuck_out_tongue_closed_eyes:

julianneswinoga (approved):
Squash before merge!

PR #98, opened 2021-05-15
Fix southern hemisphere returning DEC <-90

trishmapow:
As discussed on Discord with @ClutchplateDude. Was previously receiving nonsense DEC values of -135 that would crash OATControl. Now both target and current DEC displays work as expected.
Changes only affect southern hemisphere.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,6 @@
+**V1.9.12 - Updates**
+- Fix southern hemisphere returning incorrect DEC values
+
 **V1.9.11 - Updates**
 - Add support for focuser on E1 motor for MKS board. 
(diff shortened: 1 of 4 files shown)


trishmapow:
Forgot to mention, also tested and works fine with ASCOM.
(Contrary to what I said before, the bug is visible in ASCOM but it didn't cause a hard crash so I didn't notice)

julianneswinoga (dismissed):
You need to update the version and changelog to pass the checks, but looks good to me :+1:

trishmapow:
Cool, done. Not sure what that dismiss stuff means...

julianneswinoga (approved):

PR #99, opened 2021-05-15
Better and more robust CI version check

julianneswinoga:
New `version_check.py` script that reads/parses `Version.h` and `Changelog.md`:
- Makes sure that the versions match
- Makes sure that the version has been incremented vs `origin/develop`
- Makes sure that `Changelog.md` has *some* sort of description and not just a new version header

Just seems a bit more robust than the current actions that check if the files are changed in any way.


version_check.py
```diff
@@ -0,0 +1,119 @@
...
+def die(*args, error_code=1, **kwargs):
+    print('ERROR: ', end='', file=sys.stderr)
+    print(*args, **kwargs, file=sys.stderr)
+    exit(error_code)
+
+
+class SemVerWithVPrefix(semver.VersionInfo):
+    @classmethod
+    def parse(cls, version):
+        if not version[0] in ('v', 'V'):
```
andre-stefanov: lets allow only uppercase

version_check.py
```diff
@@ -0,0 +1,119 @@
...
+            raise ValueError(f"{version}: not a valid semantic version tag. Must start with 'v' or 'V'")
+        self = super(SemVerWithVPrefix, cls).parse(version[1:])
+        return self
+
+    def __str__(self):
+        return 'v' + super(SemVerWithVPrefix, self).__str__()
```
andre-stefanov: definitelly uppercase V

.github/workflows/changelog.yml
```diff
@@ -1,34 +1,20 @@
...
-          if [[ $(git diff origin/develop Changelog.md) ]]; then
-            echo "Version was changed"
-            exit 0 
-          else
-            echo "No version change detected"
-            exit 1 
-          fi
+          python3 -m pip install --upgrade pip
+          python3 -m pip install -r requirements_version_check.txt
+      - name: Run PlatformIO
```
andre-stefanov: rename the task name since it has nothing to do with pio

.github/workflows/changelog.yml
```diff
@@ -1,34 +1,20 @@
...
+        run: python3 version_check.py
```
andre-stefanov, line 20: pass the target branch to the script. this way we can use this script even for other branches (e.g. future v2.x.x branch)

version_check.py
```diff
@@ -0,0 +1,118 @@
...
+        version_str = next(v.strip('"') for k, v in version_defines.items() if 'version' in k.lower())
+    except StopIteration:
+        die(f'Could not find version define in\n{version_header_contents}: {version_defines}')
+    try:
+        version = SemVerWithVPrefix.parse(version_str)
+    except ValueError as e:
+        die(f'SemVer error: {e}')
+
+    print(f'Version is {version}')
+    return version
```
andre-stefanov, line 63: put this into the first try

version_check.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def get_header_version(version_header_contents: str) -> semver.VersionInfo:
+    version_defines = get_header_defines(version_header_contents)
+    try:
+        version_str = next(v.strip('"') for k, v in version_defines.items() if 'version' in k.lower())
+    except StopIteration:
+        die(f'Could not find version define in\n{version_header_contents}: {version_defines}')
+    try:
+        version = SemVerWithVPrefix.parse(version_str)
```
andre-stefanov, line 58: put this into the first try. the flow will be terminated with `die` but if this exit function changes some day, this code here will break because version_str is not initialized yet

version_check.py
```diff
@@ -0,0 +1,118 @@
...
+        return 'v' + super(SemVerWithVPrefix, self).__str__()
+
+
+def get_header_defines(header_contents: str) -> Dict[str, str]:
+    header = CppHeaderParser.CppHeader(header_contents, argType='string')
+    defines_dict = {}
+    for d in header.defines_detail:
+        define_tup = d['value'].split(' ', maxsplit=1)
+        defines_dict[define_tup[0]] = define_tup[1]
+    return defines_dict
```
andre-stefanov: can be replaced with 
`return dict(map(lambda d: d['value'].split(' ', maxsplit=1), header.defines_detail))`

version_check.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def find_child_text(node) -> Optional[str]:
```
andre-stefanov: node parameter type missing `find_child_text(node: dict)`

version_check.py
```diff
@@ -0,0 +1,118 @@
...
+    if 'children' in node.keys():
+        if isinstance(node['children'], list):
+            return find_child_text(node['children'][0])
+    elif 'text' in node.keys():
+        return node['text']
+    return None
```
andre-stefanov: this code can fail if e.g. 'children' key exists and its value is an empty list. also you can simply use node.get('children') to get the value or None if key does not exist. Assuming that 'children' value is always a list (if existent), the code can be shortened to:

```python
def find_child_text(node: dict) -> Optional[str]:
    if node.get('children'):
        return find_child_text(node.get('children')[0])
    else:
        return node.get('text')
```

andre-stefanov (changes requested):

version_check.py
julianneswinoga: Mmm I don't agree. I want the try/except to explicitly catch exactly the type of error that I'm expecting. If I change it to something like:
```python

    try:
        version_str = next(v.strip('"') for k, v in version_defines.items() if 'version' in k.lower())
        version = SemVerWithVPrefix.parse(version_str)
    except StopIteration:
        die(f'Could not find version define in\n{version_header_contents}: {version_defines}')
    except ValueError as e:
        die(f'SemVer error: {e}')
```
I don't want the first line to raise a `ValueError`, which was intended to catch a semver error. If the first line raises anything except for a `StopIteration` I **want** the script to exit non-peacefully (i.e. with a stack trace)

version_check.py
julianneswinoga: I don't like the implicit assumption that `['children']` always contains a list. Mistune documentation is pretty sparse on it's AST generation (https://mistune.readthedocs.io/en/latest/guide.html?highlight=ast#astrenderer). I'll refactor a bit to fix the `[]` bug, maybe you'll like the new version better :smiley:

andre-stefanov (approved):

PR #109, opened 2021-06-06
Support Digital level with software I2C

ClutchplateDude:


src/Gyro.cpp
```diff
@@ -5,27 +5,35 @@
...
+#endif
 POP_NO_WARNINGS
 
-/**
- * Tilt, roll, and temperature measurementusing the MPU-6050 MEMS gyro.
- * See: https://invensense.tdk.com/products/motion-tracking/6-axis/mpu-6050/
- * Datasheet: https://invensense.tdk.com/wp-content/uploads/2015/02/MPU-6000-Datasheet1.pdf
- * Register descriptions: https://invensense.tdk.com/wp-content/uploads/2015/02/MPU-6000-Register-Map1.pdf
- * */
+// Tilt, roll, and temperature measurementusing the MPU-6050 MEMS gyro.
```
julianneswinoga, line 16: Try not to change comment styles with no additional changes, it just adds code churn

src/Gyro.cpp
```diff
@@ -5,27 +5,35 @@
...
+// See: https://invensense.tdk.com/products/motion-tracking/6-axis/mpu-6050/
+// Datasheet: https://invensense.tdk.com/wp-content/uploads/2015/02/MPU-6000-Datasheet1.pdf
+// Register descriptions: https://invensense.tdk.com/wp-content/uploads/2015/02/MPU-6000-Register-Map1.pdf
 
-bool Gyro::isPresent(false);
+// Static var to keep track of whether the MPU6050 was found
+bool Gyro::isPresent = false;
 
+// Starts up the MPU-6050 device.
```
julianneswinoga, line 24: Code churn here too

src/Gyro.cpp
```diff
@@ -34,67 +42,88 @@ void Gyro::startup()
     Wire.write(MPU6050_REG_WHO_AM_I);
     Wire.endTransmission(true);
     Wire.requestFrom(MPU6050_I2C_ADDR, 1, 1);
-    byte id = (Wire.read() >> 1) & 0x3F;    
+    byte id = Wire.read();
```
julianneswinoga:
```C++
const byte id = ((Wire.read() >> 1) & 0x3F);
isPresent = (id == 0x34);
```
?
There are wayyyy better ways to do register reading/comparisons but this can still be cleaned up a tad

src/Gyro.cpp
```diff
@@ -34,67 +42,88 @@ void Gyro::startup()
...
-    // Execute 1 byte write to MPU6050_REG_PWR_MGMT_1
+    // Execute 1 byte write to MPU6050_REG_PWR_MGMT_1 to disable sleep and set 8MHz clock
     Wire.beginTransmission(MPU6050_I2C_ADDR);
     Wire.write(MPU6050_REG_PWR_MGMT_1);
-    Wire.write(0);      // Disable sleep, 8 MHz clock
+    Wire.write(0); // Disable sleep, 8 MHz clock
     Wire.endTransmission(true);
 
-    // Execute 1 byte write to MPU6050_REG_PWR_MGMT_1
+    // Execute 1 byte read/write to MPU6050_REG_ACCEL_CONFIG to set 4g sensititvity
```
julianneswinoga: Spelling `sensitivity`

src/Gyro.cpp
```diff
@@ -34,67 +42,88 @@ void Gyro::startup()
...
+    Wire.beginTransmission(MPU6050_I2C_ADDR);
+    Wire.write(MPU6050_REG_ACCEL_CONFIG);
+    Wire.endTransmission();
+    Wire.requestFrom(MPU6050_I2C_ADDR, 1);
+    byte x = Wire.read(); //the value of Register-28 is in x
+    x = (x & 0b11100111);     //clear values of Bit4 and Bit3 fo 2g sensitivity
```
julianneswinoga: You say 4g sensitivity above, which one is it?

src/Gyro.cpp
```diff
@@ -34,67 +42,88 @@ void Gyro::startup()
...
+    Wire.write(0); // Disable sleep, 8 MHz clock
     Wire.endTransmission(true);
 
-    // Execute 1 byte write to MPU6050_REG_PWR_MGMT_1
+    // Execute 1 byte read/write to MPU6050_REG_ACCEL_CONFIG to set 4g sensititvity
+    Wire.beginTransmission(MPU6050_I2C_ADDR);
+    Wire.write(MPU6050_REG_ACCEL_CONFIG);
+    Wire.endTransmission();
+    Wire.requestFrom(MPU6050_I2C_ADDR, 1);
+    byte x = Wire.read(); //the value of Register-28 is in x
```
julianneswinoga:
```C++
const byte accelCfgReg = (Wire.read() & 0b11100111);
```
?

src/Gyro.cpp
```diff
@@ -34,67 +42,88 @@ void Gyro::startup()
...
+
+        //LOGV7(DEBUG_INFO, F("GYRO: Values : %d, %d   %d, %d   %d, %d"), b0,b1,b2,b3,b4,b5);
 
         // Calculating the Pitch angle (rotation around Y-axis)
         result.pitchAngle += ((atanf(-1 * AcX / sqrtf(powf(AcY, 2) + powf(AcZ, 2))) * 180.0f / static_cast<float>(PI)) * 2.0f) / 2.0f;
         // Calculating the Roll angle (rotation around X-axis)
         result.rollAngle += ((atanf(-1 * AcY / sqrtf(powf(AcX, 2) + powf(AcZ, 2))) * 180.0f / static_cast<float>(PI)) * 2.0f) / 2.0f;
 
-        delay(10);  // Decorrelate measurements
+        delay(5); // Decorrelate measurements
```
julianneswinoga: Why change the update rate + window size here? In the initialization you set 5Hz bandwidth but here you're checking at 200Hz (1/(5ms) == 200Hz). Am I missing something?

src/Gyro.cpp
```diff
@@ -34,67 +42,88 @@ void Gyro::startup()
...
 
     for (int i = 0; i < windowSize; i++)
     {
         // Execute 6 byte read from MPU6050_REG_WHO_AM_I
         Wire.beginTransmission(MPU6050_I2C_ADDR);
-        Wire.write(MPU6050_REG_ACCEL_XOUT_H); 
+        Wire.write(MPU6050_REG_ACCEL_XOUT_H);
         Wire.endTransmission(false);
-        Wire.requestFrom(MPU6050_I2C_ADDR, 6, 1);  // Read 6 registers total, each axis value is stored in 2 registers
-        int16_t AcX = Wire.read() << 8 | Wire.read(); // X-axis value
```
julianneswinoga, line 88: Mmm yeah this was sketchy, gj breaking it up :+1:

src/Gyro.cpp
```diff
@@ -34,67 +42,88 @@ void Gyro::startup()
...
-        int16_t AcY = Wire.read() << 8 | Wire.read(); // Y-axis value
-        int16_t AcZ = Wire.read() << 8 | Wire.read(); // Z-axis value
+        Wire.requestFrom(MPU6050_I2C_ADDR, 6, 1);     // Read 6 registers total, each axis value is stored in 2 registers
+        uint8_t b0 =  Wire.read();
```
julianneswinoga: I'd rename to `AxXLow`/`AcXHigh` or something to be more explicit, but up to you. (I would also pepper `const` everywhere but that's just me :grin: )

julianneswinoga (changes requested):
@andre-stefanov @ClutchplateDude @CTetford 
In the future we should look into using git submodules instead of raw source clones... It would guarantee the git hash and be easy to update.

src/Gyro.cpp
ClutchplateDude: You know how you say always leave code better than how you find it? enums vs. typedefs, etc. coding style is one of those. :-)

src/Gyro.cpp
ClutchplateDude: This used to be on one line, but I split it for debugging purposes.... is there a request here?

src/Gyro.cpp
ClutchplateDude: Fixed

src/Gyro.cpp
ClutchplateDude: Fixed

src/Gyro.cpp
ClutchplateDude: Yup.

src/Gyro.cpp
ClutchplateDude: Not sure. Maybe I'm not understanding how this works.

src/Gyro.cpp
julianneswinoga: Yeah I know, it's a fine line :)
My reasoning:
- There are no functional benefits
- It's subjective (for now, until we have coding guidelines)

src/Gyro.cpp
julianneswinoga: In my opinion it's a much cleaner implementation, shows the intent a lot better, const-correctness, etc. Up to you in the end though :upside_down_face:

src/Gyro.cpp
ClutchplateDude: Yeah, no guarantee in what order the compiler calls read

src/Gyro.cpp
ClutchplateDude: Fixed

src/Gyro.cpp
ClutchplateDude: The style is pretty consistent (with expections), but nowhere do we use this style of commenting...

src/Gyro.cpp
ClutchplateDude: Oh, I didn't realize that was new code you wrote, lol. Like I said while trying to get this running, I wanted to see what actually was read before the bit fiddling.. But now that it works, I'll change it to the more optimized now.

src/Gyro.cpp
julianneswinoga: I meant the code line, same as the other register refactor above

src/Gyro.cpp
ClutchplateDude: Fixed.

boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
```diff
@@ -150,6 +150,8 @@
   #define ALT_DRIVER_ADDRESS 0b00
 #endif
 
+// Digital Level addon - not officially supported on ESP32. If you find two free pins, add GYRO_SOFTWARE_SCL_PIN  and GYRO_SOFTWARE_SDA_PIN to your local config
```
julianneswinoga: Shouldn't this be MKS GenL V1 instead of ESP32?

src/Gyro.cpp
```diff
@@ -5,100 +5,133 @@
...
+    Wire.write(0); // Disable sleep, 8 MHz clock
     Wire.endTransmission(true);
 
-    // Execute 1 byte write to MPU6050_REG_PWR_MGMT_1
+    // Execute 1 byte read/write to MPU6050_REG_ACCEL_CONFIG to set 2g sensitivity
+    Wire.beginTransmission(MPU6050_I2C_ADDR);
+    Wire.write(MPU6050_REG_ACCEL_CONFIG);
+    Wire.endTransmission();
+    Wire.requestFrom(MPU6050_I2C_ADDR, 1);
+    byte accelConfigReg = Wire.read();              //the value of Register-28 is in x
```
julianneswinoga: Could you do the same thing here as I requested above?
```C++
const byte accelConfigReg = (Wire.read() & 0b11100111);
```

src/Gyro.cpp
```diff
@@ -5,100 +5,133 @@
...
+    const int16_t accelInZ = accelZMSByte << 8 | accelZLSByte; // Z-axis value
+
+    // Calculating the Pitch angle (rotation around Y-axis)
+    result.pitchAngle += ((atanf(-1 * accelInX / sqrtf(powf(accelInY, 2) + powf(accelInZ, 2))) * 180.0f / static_cast<float>(PI)) * 2.0f) / 2.0f;
+
+    // Calculating the Roll angle (rotation around X-axis)
+    result.rollAngle += ((atanf(-1 * accelInY / sqrtf(powf(accelInX, 2) + powf(accelInZ, 2))) * 180.0f / static_cast<float>(PI)) * 2.0f) / 2.0f;
+
+#if USE_GYRO_WITH_SOFTWAREI2C == 0
+    // if (Wire.getWireTimeoutFlag())
```
julianneswinoga, line 133: Delete unused debug code

src/Gyro.cpp
```diff
@@ -5,100 +5,133 @@
...
+    LOGV1(DEBUG_INFO, F("GYRO:: Starting software I2C for MPU6050 comms."));
+#else
+    LOGV1(DEBUG_INFO, F("GYRO:: Starting hardware I2C for MPU6050 comms."));
+#endif
+
     Wire.begin();
+    Wire.setClock(100000); // Set lowest clock speed (100kHz) to make the communication more reliable
 
+#if USE_GYRO_WITH_SOFTWAREI2C == 0
+    // Wire.setWireTimeout(3000, true); // Set I2C timeout if the Wire.h calls are hanging due to a HW issue
```
julianneswinoga, line 41: Delete unused debug code

src/c76_menuCAL.hpp
```diff
@@ -1231,28 +1248,46 @@ void printCalibrationSubmenu()
...
-    sprintf(scratchBuffer, "R: -------------");
-    makeIndicator(scratchBuffer, angles.rollAngle - mount.getRollCalibrationAngle());
-    lcdMenu.printMenu(scratchBuffer);
+    accumulator.rollAngle += angles.rollAngle;
+    if (calCount >= ACCUMULATOR_COUNT)
+    {
+      accumulator.rollAngle /= (float)ACCUMULATOR_COUNT;
+      sprintf(scratchBuffer, "R: ");
+      makeIndicator(scratchBuffer, accumulator.rollAngle - mount.getRollCalibrationAngle());
+      scratchBuffer[16] = 0;
```
julianneswinoga: Use `'\0'` instead of `0` for null byte in string

src/c76_menuCAL.hpp
```diff
@@ -1231,28 +1248,46 @@ void printCalibrationSubmenu()
...
   {
+    calCount++;
     auto angles = Gyro::getCurrentAngles();
-    sprintf(scratchBuffer, "R: -------------");
-    makeIndicator(scratchBuffer, angles.rollAngle - mount.getRollCalibrationAngle());
-    lcdMenu.printMenu(scratchBuffer);
+    accumulator.rollAngle += angles.rollAngle;
+    if (calCount >= ACCUMULATOR_COUNT)
+    {
+      accumulator.rollAngle /= (float)ACCUMULATOR_COUNT;
```
julianneswinoga: Could you instead implement the details in the gyro module instead of the menu file? It doesn't make sense to do this here.

Also while we're changing this, doing this dance with a calibration counter, accumulator, reset, etc seems like a very simplistic way to implement a low pass filter. Something like an exponential low pass filter ([example C implementation](https://kiritchatterjee.wordpress.com/2014/11/10/a-simple-digital-low-pass-filter-in-c/)) is very simple to implement and would work fantastic for our use case. :smile: 
I can definitely help you out here if you need it!

julianneswinoga (changes requested):

src/Gyro.cpp
ClutchplateDude: This is not debug code. There is a library that supports timeouts

src/Gyro.cpp
ClutchplateDude: Ditto

src/c76_menuCAL.hpp
ClutchplateDude: Fixed.

src/c76_menuCAL.hpp
ClutchplateDude: I don't want the Gyro class to do this, it would make it unneccesarily slower. But here the frequent LCD updates make it unreadable. I want it a lot slower, but only for the LCD. So this is the right place. Won't fix.

ClutchplateDude: OK, that low pass filter works really well (in the test app). Trying to see if I can get it to behave in the firmware

src/Gyro.cpp
ClutchplateDude: Done

src/Gyro.cpp
```diff
@@ -5,100 +5,133 @@
...
+    Wire.write(MPU6050_REG_ACCEL_CONFIG);
+    Wire.write(accelConfigReg);
+    Wire.endTransmission();
+
+    // Execute 1 byte write to MPU6050_REG_CONFIG to set slowest refresh rate
     Wire.beginTransmission(MPU6050_I2C_ADDR);
     Wire.write(MPU6050_REG_CONFIG);
-    Wire.write(6);      // 5Hz bandwidth (lowest) for smoothing
+    Wire.write(6); // 5Hz bandwidth (lowest) for smoothest readings
     Wire.endTransmission(true);
```
andre-stefanov, line 78: you could bundle these calls to a function something like `send(reg, val)`

andre-stefanov: even with default value `null` for val if only register has to be sent

src/Gyro.cpp
julianneswinoga: Why is it commented out then? As far as I'm concerned, if code is commented out it's debug.

src/c76_menuCAL.hpp
```diff
@@ -1232,27 +1249,36 @@ void printCalibrationSubmenu()
   else if (calState == ROLL_OFFSET_CALIBRATION)
   {
     auto angles = Gyro::getCurrentAngles();
-    sprintf(scratchBuffer, "R: -------------");
-    makeIndicator(scratchBuffer, angles.rollAngle - mount.getRollCalibrationAngle());
-    lcdMenu.printMenu(scratchBuffer);
+    float delta = accumulator.rollAngle - angles.rollAngle;
```
julianneswinoga, line 1252: I still think that this is not the place for the calculations, they should be abstracted in the gyro interface such that the menu code doesn't care about how the calculations work. The menu just wants the angles and the gyro takes care of how that happens, which reduces module coupling (https://en.wikipedia.org/wiki/Coupling_(computer_programming))

src/c76_menuCAL.hpp
```diff
@@ -94,6 +94,12 @@ enum
 // Start off with Polar Alignment higlighted.
 byte calState = HIGHLIGHT_FIRST;
 
+// Variables to accumulate and smooth the accelerometer readings of the digital level
+#if USE_GYRO_LEVEL == 1
+angle_t accumulator;
+float LPFBeta = 0.6f;
```
julianneswinoga, line 100: You should get in the habit of using `const` wherever possible: https://isocpp.org/wiki/faq/const-correctness

julianneswinoga (changes requested):

src/Gyro.cpp
ClutchplateDude: There is a library upgrade coming where this will be used. I don't want to have to think about getting this code back once it's included.

src/c76_menuCAL.hpp
ClutchplateDude: I disagree. As a consumer of the class, I might not want it smoothed. I might want a different smoothing factor. I might be calling it irregularly and smoothing would be worse....

src/c76_menuCAL.hpp
ClutchplateDude: Sure.

src/c76_menuCAL.hpp
andre-stefanov: Then it should be a parameter on gyro api. But here menu is changing actual behaviour while it should only trigger a feature without knowing the actual implementation.

Imagine we want to provide this as meade as well.

julianneswinoga: > As a consumer of the class, I might not want it smoothed. I might want a different smoothing factor.

Simple, two functions then: `Gyro::getCurrentAnglesRaw(void)` and `Gyro::getCurrentAnglesFiltered(float smoothFactor)`

> I might be calling it irregularly and smoothing would be worse

That's up to the consumer of the class to use the interface correctly. You can fix that in the gyro interface with more complexity (threading, etc) but that's not needed for simple usage like this.

src/Gyro.cpp
julianneswinoga: Ideally we would implement that code when the functionality is needed, but I can understand why you want to leave it now. Can you at least add a `// TODO` comment explaining why the code is unused?

ClutchplateDude:
Abandoning this PR. Digital Level is very underused and code change is too old.

PR #112, opened 2021-06-07
Added installation script for windows hosts

andre-stefanov:
You have to add board.txt containing the pio environment to the root directory in order for this script to work

what it does:
- checks for python installation and redirects the user to the microsoft store python page if python is missing
- installs platformio
- builds firmware
- uploads firmware to the board

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -1,0 +1,35 @@
+@echo off
+
+:: change current directory to the one where this script is located
+:: if executed as administrator it would change to system32 otherwise
+pushd %~dp0
+
+python --version 2>NUL
+if errorlevel 1 goto errorNoPython
+
+if exist %UserProfile%\.platformio\penv\Scripts\ (
+  echo PIO detected. Skip installation...
+) else (
+  python get-platformio.py
+)
+
+SET PATH=%PATH%;%UserProfile%\.platformio\penv\Scripts
+
+if exist board.txt (
+  set /p PIO_ENVIRONMENT=<board.txt
+) else (
+  echo board.txt not found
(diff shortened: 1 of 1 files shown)


install.bat
```diff
@@ -0,0 +1,35 @@
...
+  pause
+  exit
+)
+
+pio run -e %PIO_ENVIRONMENT% -t upload
+
+pause
+exit
+
+:errorNoPython
```
julianneswinoga, line 31: Is it possible to install python through this script? The less complicated things a user has to do here the better.

install.bat
```diff
@@ -0,0 +1,35 @@
+@echo off
+
+:: change current directory to the one where this script is located
+:: if executed as administrator it would change to system32 otherwise
+pushd %~dp0
+
+python --version 2>NUL
+if errorlevel 1 goto errorNoPython
+
+if exist %UserProfile%\.platformio\penv\Scripts\ (
```
julianneswinoga, line 10: Can we not check if `pio` is present in the path? Is this the best way to check for it?

install.bat
```diff
@@ -0,0 +1,35 @@
...
+  echo PIO detected. Skip installation...
+) else (
+  python get-platformio.py
+)
+
+SET PATH=%PATH%;%UserProfile%\.platformio\penv\Scripts
+
+if exist board.txt (
```
julianneswinoga, line 18: What's the reason the normal `platformio.ini` doesn't work?

julianneswinoga (changes requested):

PR #114, opened 2021-06-11
Add formatting check

julianneswinoga:
Style was loosely based off of the Microsoft style (started with `clang-format --style=Microsoft --dump-config > .clang-format` and made changes)
Look over your favorite files and let me know if you'd like to propose any style changes @OpenAstroTech/developer 's :smile: 

See #111 for work-in-progress PR, as well as what @openastrotech-bot would comment if the check fails


julianneswinoga:
The reason that the check doesn't run on this PR is because of how `pull_request_target` works (only runs if that workflow file has not been modified by the PR), and we need that if we want the workflow to have read/write permissions, which we need in order to tell the bot to add a comment if the check fails :sweat_smile:

Configuration_sample_local.hpp
```diff
@@ -27,33 +27,36 @@
...
  */
-#define RA_DRIVER_TYPE      DRIVER_TYPE_ULN2003
-#define DEC_DRIVER_TYPE     DRIVER_TYPE_ULN2003
-#define RA_STEPPER_SPEED          400   // Max. Speed = 600 for 28BYJ-48 and 3000 for NEMA17. Defaults = 400 for 28BYJ-48 and 1200 for NEMA17
-#define RA_STEPPER_ACCELERATION   600   // Defaults: 600 for 28BYJ-48, 6000 for NEMA17
-#define DEC_STEPPER_SPEED          600   // Max. Speed = 600 for 28BYJ-48 and 3000 for NEMA17. Defaults = 600 for 28BYJ-48 and 1300 for NEMA17
-#define DEC_STEPPER_ACCELERATION   600   // Defaults: 600 for 28BYJ-48, 6000 for NEMA17
+#define RA_DRIVER_TYPE DRIVER_TYPE_ULN2003
+#define DEC_DRIVER_TYPE DRIVER_TYPE_ULN2003
+#define RA_STEPPER_SPEED                                                                                               \
```
ClutchplateDude: What is this mess it's doing? Splitting it over two lines? No more lining up values? Booo, this needs to change

src/DayTime.cpp
```diff
@@ -14,337 +14,306 @@
...
-
-  // Third digit?
-  if ((s[i] >= '0') && (s[i] <= '9'))
-  {
+    DayTime result;
+    int i    = 0;
+    long sgn = 1;
+    LOGV2(DEBUG_MEADE, F("DayTime: Parse Coord from [%s]"), s.c_str());
+    // Check whether we have a sign. This should be able to parse RA and DEC strings (RA never has a sign, and DEC should always have one).
+    if ((s[i] == '-') || (s[i] == '+')) {
```
ClutchplateDude: Opeing brace on seperate new line, please.

src/EPROMStore.cpp
```diff
@@ -156,161 +158,168 @@ void EEPROMStore::displayContents()
...
 
-  unsigned check = (MAGIC_MARKER_MASK | item);
-  unsigned result=(MAGIC_MARKER_VALUE | item);
-  bool res = ((marker & check) == result);
-  LOGV6(DEBUG_EEPROM, F("EEPROM: IsDataPresent (%x). Checking (%x & %x) == %x => %d"), item, marker, check, result, res);
+    unsigned check  = (MAGIC_MARKER_MASK | item);
+    unsigned result = (MAGIC_MARKER_VALUE | item);
+    bool res        = ((marker & check) == result);
+    LOGV6(DEBUG_EEPROM,
+          F("EEPROM: IsDataPresent (%x). Checking (%x & %x) == %x => %d"),
```
ClutchplateDude: Urgh. What makes this go over multiple lines? Can we make it stay on one line?

src/EPROMStore.cpp
```diff
@@ -319,450 +328,424 @@ void EEPROMStore::updateFlagsExtended(ExtendedItemFlag item)
...
-  // There is no item flag for brightness - it is assumed to always be present
+    // There is no item flag for brightness - it is assumed to always be present
 
-  byte brightness = readUint8(LCD_BRIGHTNESS_ADDR);
-  if (brightness == 0)
-    brightness = 10; // Have a reasonable minimum in case nothing is stored
-  return brightness; // dimensionless scalar
+    byte brightness = readUint8(LCD_BRIGHTNESS_ADDR);
+    if (brightness == 0)
+        brightness = 10;  // Have a reasonable minimum in case nothing is stored
```
ClutchplateDude, line 376: Is there a rule available that enforces braces on all if()s?

src/EPROMStore.hpp
```diff
@@ -8,164 +8,188 @@
...
+        _LONGITUDE_ADDR_1           = 13,  // Int16
+        LCD_BRIGHTNESS_ADDR         = 16,  // Uint8
+        PITCH_OFFSET_ADDR           = 17,
+        _PITCH_OFFSET_ADDR_1        = 18,  // Uint16
+        ROLL_OFFSET_ADDR            = 19,
+        _ROLL_OFFSET_ADDR_1         = 20,  // Uint16
+        EXTENDED_FLAGS_ADDR         = 21,
+        _EXTENDED_FLAGS_ADDR_1      = 22,  // Uint16
+        RA_PARKING_POS_ADDR         = 23,
+        _RA_PARKING_POS_ADDR_1,
```
ClutchplateDude, line 147: Probably should add the explicit numbers here, so they format better.

.clang-format
```diff
@@ -0,0 +1,148 @@
+---
+Language:        Cpp
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
```
ClutchplateDude: I think this is the one that should be true.

.clang-format
```diff
@@ -0,0 +1,148 @@
...
+BreakBeforeBinaryOperators: All
+BreakBeforeBraces: Custom
+BreakBeforeInheritanceComma: false
+BreakInheritanceList: BeforeColon
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     120
```
ClutchplateDude: Let's up this to 150

src/c65_startup.hpp
```diff
@@ -28,178 +29,171 @@ enum startupState_t {
...
-      }
-    }
-    break;
+    } break;
 
-    #if USE_GYRO_LEVEL == 1
-    case StartupSetRoll : {
-      inStartup = false;
-      LOGV1(DEBUG_INFO, F("STARTUP: Switching to CAL menu!"));
+        #if USE_GYRO_LEVEL == 1
```
ClutchplateDude, line 120: This is a weird indentation. And in some files #if and #ifdefs don't indent at all?

.clang-format
```diff
@@ -0,0 +1,148 @@
...
+    SortPriority:    0
+  - Regex:           '^(<|"(gtest|gmock|isl|json)/)'
+    Priority:        3
+    SortPriority:    0
+  - Regex:           '.*'
+    Priority:        1
+    SortPriority:    0
+IncludeIsMainRegex: '(Test)?$'
+IncludeIsMainSourceRegex: ''
+IndentCaseLabels: false
```
ClutchplateDude: This should be true

ClutchplateDude (commented):
Looked over some of the files to see what it's doing.... needs some tweaks, but I like having it be consistent over the code base! Thanks.

Configuration_sample_local.hpp
julianneswinoga: Hmm, yeah this seems to be because of the inline comment

src/DayTime.cpp
julianneswinoga: I vastly prefer the opening brace for control flow to be on the same line, but I think we should have a team vote for this point :grin:

src/EPROMStore.cpp
julianneswinoga: It splits the lines up because of the character limit (120), lets check back on this when we bump that up

.clang-format
julianneswinoga: I think 150 is a bit much, but yeah the 'standard' 120 is a bit too restrictive I think. 140?

src/EPROMStore.cpp
julianneswinoga: No, `clang-format` won't alter the code like this but `clang-tidy` has that option. We could look at some `clang-tidy` integration in the future!

src/EPROMStore.hpp
julianneswinoga: Yeah, or rearrange the enum such that they're in the correct order. We could also add some `static_assert()`s to double check that no-one breaks the order unintentionally

src/c65_startup.hpp
julianneswinoga: What's the weird indentation? The preprocessor instructions are indented separately than the actual code

src/EPROMStore.cpp
ClutchplateDude: I don't mean for it to change it, just to flag it.

.clang-format
julianneswinoga: I don't agree, in my mind that leads to way too much indentation cruft since the case body is always indented 1 more than the case label.

src/DayTime.cpp
```diff
@@ -14,337 +14,331 @@
...
-      if (inMacro)
-      {
-        switch (macro)
-        {
-        case 'd':
+        switch (*f)
         {
-          strcpy(p, achDegs);
-          p += strlen(achDegs);
+        case '{': {
```
ClutchplateDude: This opening curly needs to be on a new line.

src/c65_startup.hpp
ClutchplateDude: Oh. Interesting.

ClutchplateDude (changes requested):

.clang-format
ClutchplateDude: Sure

ClutchplateDude (approved):

PR #117, opened 2021-06-17
Fix southern hemisphere -12hr RA offset

trishmapow:
RA was being flipped the wrong way which did not affect slewing, but would display the mount upside down in software.

I see there are other PRs ready to go, so maybe I should wait to update the version and changelog?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,5 @@
 **V1.9.17 - Updates**
-- Fix southern hemisphere -12hr RA offset
+- Fix southern hemisphere returning RA offset by -12hr
 
 **V1.9.16 - Updates**
 - Add Meade extension command to move steppers by steps.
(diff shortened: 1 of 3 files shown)


julianneswinoga (dismissed):
@trishmapow Nah it's alight, smaller PRs are easier to review so they usually get through quicker.

Changelog.md
```diff
@@ -1,3 +1,6 @@
+**V1.9.17 - Updates**
+- Fix southern hemisphere -12hr RA offset
```
julianneswinoga: Could you change to clarify that this is just a reporting issue and not a slewing one?

julianneswinoga (changes requested):

Changelog.md
trishmapow: Hopefully it's clearer now

julianneswinoga (approved):

PR #118, opened 2021-06-17
Small guiding improvements

intercipere:
1. Speeds for TRK and GUIDE stepper instances should not exceed the maximum achievable speeds of accelstepper
2. Assuming a trackingspeed of 10 for simplicity sake, and a RA_PULSE_MULTIPLIER of e.g. 1.1, West speed should be 11 and East speed should be 9, not 21 W / 1 E. This also caused the East/West guidespeeds at the default multiplier of 1.5 to be wrong, at 25W / 5E. It must be 15W / 5E

### DIFF
--- a/src/Mount.cpp
+++ b/src/Mount.cpp
@@ -1412,8 +1412,8 @@ void Mount::guidePulse(byte direction, int duration) {
 
     case EAST:
     // We were in tracking mode before guiding, so no need to update microstepping mode on RA driver
-    LOGV2(DEBUG_STEPPERS, F("STEP-guidePulse:  TRK.setSpeed(%f)"), (raGuidingSpeed * (-1.0f * RA_PULSE_MULTIPLIER + 1) + raGuidingSpeed));
-    _stepperTRK->setSpeed(raGuidingSpeed * (-1.0f * RA_PULSE_MULTIPLIER + 1) + raGuidingSpeed);   // Slower than siderael
+    LOGV2(DEBUG_STEPPERS, F("STEP-guidePulse:  TRK.setSpeed(%f)"), (raGuidingSpeed * (2.0f - RA_PULSE_MULTIPLIER)));
+    _stepperTRK->setSpeed(raGuidingSpeed * (2.0f - RA_PULSE_MULTIPLIER));   // Slower than siderael
     _mountStatus |= STATUS_GUIDE_PULSE | STATUS_GUIDE_PULSE_RA;
     _guideRaEndTime = millis() + duration;
     break;
--- a/src/Mount.cpp
+++ b/src/Mount.cpp
@@ -245,8 +245,8 @@ void Mount::configureRAStepper(byte pin1, byte pin2, int maxSpeed, int maxAccele
   // Use another AccelStepper to run the RA motor as well. This instance tracks earths rotation.
   _stepperTRK = new AccelStepper(AccelStepper::DRIVER, pin1, pin2);
 
-  _stepperTRK->setMaxSpeed(10000);
-  _stepperTRK->setAcceleration(5000);
+  _stepperTRK->setMaxSpeed(2000);
+  _stepperTRK->setAcceleration(15000);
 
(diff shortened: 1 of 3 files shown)


src/Mount.cpp
```diff
@@ -1404,16 +1404,16 @@ void Mount::guidePulse(byte direction, int duration) {
...
     _mountStatus |= STATUS_GUIDE_PULSE | STATUS_GUIDE_PULSE_RA;
     _guideRaEndTime = millis() + duration;
     break;
 
     case EAST:
     // We were in tracking mode before guiding, so no need to update microstepping mode on RA driver
-    LOGV2(DEBUG_STEPPERS, F("STEP-guidePulse:  TRK.setSpeed(%f)"), (RA_PULSE_MULTIPLIER - 1) * raGuidingSpeed);
-    _stepperTRK->setSpeed((RA_PULSE_MULTIPLIER - 1) * raGuidingSpeed);   // Slower than siderael
+    LOGV2(DEBUG_STEPPERS, F("STEP-guidePulse:  TRK.setSpeed(%f)"), (raGuidingSpeed * (-1.0f * RA_PULSE_MULTIPLIER + 1) + raGuidingSpeed));
+    _stepperTRK->setSpeed(raGuidingSpeed * (-1.0f * RA_PULSE_MULTIPLIER + 1) + raGuidingSpeed);   // Slower than siderael
```
ClutchplateDude: Let's change this to 
_stepperTRK->setSpeed(raGuidingSpeed * (2.0f - RA_PULSE_MULTIPLIER ) ); // Slower than siderael

ClutchplateDude (changes requested):

src/Mount.cpp
intercipere: right, thats a little less weird maths :)

src/Mount.cpp
```diff
@@ -245,8 +245,8 @@ void Mount::configureRAStepper(byte pin1, byte pin2, int maxSpeed, int maxAccele
   // Use another AccelStepper to run the RA motor as well. This instance tracks earths rotation.
   _stepperTRK = new AccelStepper(AccelStepper::DRIVER, pin1, pin2);
 
-  _stepperTRK->setMaxSpeed(10000);
-  _stepperTRK->setAcceleration(5000);
+  _stepperTRK->setMaxSpeed(2000);
```
julianneswinoga, line 248: Is there a reference (or calculation) we can use here instead of a magic number?

intercipere: You mean the max speed of 2000? Thats the upper limit of what Accelstepper can achieve

ClutchplateDude (approved):

src/Mount.cpp
julianneswinoga: Yes, `2000` does not convey a lot of information when reading the code though. Can `2000` be derived from an existing constant or `#define`? Can `2000` be converted into more meaningful units? How was `2000` determined to be the correct number? etc.

ClutchplateDude: It's steps/second. You could make a constant like TwoThousandStepsPerSec... or add comments... or know what setSpeed() expects 😁

julianneswinoga: Ok, then would it be reasonable to convert from something like the earth's rotation rate (360.985deg/day) and back down to the local stepper frame?

ClutchplateDude: No. Its purely the MAXIMUM steps per second that we can run the steppers at. It has no bearing on the earth or anything except maybe the interrupt frequency….

julianneswinoga: So what's the benefit of changing it from 10000step/s to 2000step/s then? If it's based on our interrupt frequency we should make a concrete link to that, so that they can be kept in sync.

ClutchplateDude: There is no benefit, since we never issue speeds that high anyway.

ClutchplateDude: We could name it MAX_PHYSICAL_SPEED

PR #119, opened 2021-06-26
V1.9.19 - Updates

ClutchplateDude:
- Added support to reset the focus stepper position
- Fixed some focus validation tests

### DIFF
--- a/Configuration_adv.hpp
+++ b/Configuration_adv.hpp
@@ -447,8 +447,10 @@
 // If this is set to 0 you still have a GO menu that has Home and Park.
   #define SUPPORT_POINTS_OF_INTEREST   1
 
-// Set this to 1 to support Guided Startup 
-  #define SUPPORT_GUIDED_STARTUP       1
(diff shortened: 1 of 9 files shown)


Configuration_adv.hpp
```diff
@@ -460,7 +462,9 @@
   #define SUPPORT_INFO_DISPLAY         1
 #else
   #define SUPPORT_POINTS_OF_INTEREST 0
-  #define SUPPORT_GUIDED_STARTUP     0
+  #ifndef SUPPORT_GUIDED_STARTUP       
+    #define SUPPORT_GUIDED_STARTUP   0
+  #endif
```
andre-stefanov: If there is no display available, this change would allow the user to define `SUPPORT_GUIDED_STARTUP=1`. we should rather check if it is set, and if it is 1, throw an error

ClutchplateDude: Ahh, good point.

andre-stefanov (approved):

PR #122, opened 2021-07-16
Hall sensor homing support

ClutchplateDude:
- Support for a Hall sensor homing routine for the RA ring.


src/MeadeCommandProcessor.cpp
```diff
@@ -496,6 +496,25 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
 //      Returns:
 //        "1" if successfully scheduled
 //
+// :MHR#
+//      Description:
+//        Home RA stepper via Hall sensor
+//      Information:
+//        This attempts to find the hall sensor and to home the RA ring accordingly.
+//      Parameters:
+//        If the last letter 'R' is uppercase the mount will search east first, if it is lowercase it will search west first.
```
andre-stefanov: mh ... east and west are hemisphere dependant if i understand it correctly, clockwise/counterclockwise or left/right would be more intuitive. Especially since auto-homing does not depend on the sky orientation at all ... it is a pure mount feature.

andre-stefanov: Using `R` and `r` is a bit misleading because for me R (regardless if it is upper case or lower case) stands for reverse. Also i did not see similar command syntax before here. We should provide a dedicated character for each rotation direction. We could use something like `:MHL#` and `:MHR#`

src/MeadeCommandProcessor.cpp
```diff
@@ -496,6 +496,25 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
...
+//      Remarks:
+//        The ring is first moved 30 degrees in the initial direction. If no hall sensor is encountered, it will move 60 degrees in
+//        the opposite direction. If a hall sensor is not encountered during that slew, the homing exits with a failure code (0).
+//        If the sensor is found, it will slew to the middle position of the Hall sensor trigger range and then to the offset
+//        specified in the Home offset position (set with the ":XSHRnnnn#" command).
+//        If the RA ring is positioned such that the Hall sensor is triggered when the command is recevied, teh mount will will move
```
andre-stefanov: teh mount will will

src/MeadeCommandProcessor.cpp
```diff
@@ -782,6 +809,17 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
 //      Returns:
 //        nothing
 //
+// :XSHRnnn#
```
andre-stefanov, line 812: does this also handle negative offsets?

src/Mount.cpp
```diff
@@ -2314,6 +2325,9 @@ bool Mount::isParking() const
 /////////////////////////////////
 bool Mount::isFindingHome() const
 {
+#if USE_HALL_SENSOR_RA_AUTOHOME == 1
+    return (_homing.state == HomingState::HOMING_PIN_FINDING_START) || (_homing.state == HomingState::HOMING_PIN_FINDING_END);
+#endif
     return _mountStatus & STATUS_FINDING_HOME;
```
andre-stefanov, line 2331: #else

src/Mount.cpp
```diff
@@ -2314,6 +2325,9 @@ bool Mount::isParking() const
 /////////////////////////////////
 bool Mount::isFindingHome() const
```
andre-stefanov, line 2326: mh should we maybe just protect complete autohome code with one #if instead of multiple single lines? would be easier to maintain and read

andre-stefanov (changes requested):

src/MeadeCommandProcessor.cpp
ClutchplateDude: Good point.

ClutchplateDude: Agreed

src/MeadeCommandProcessor.cpp
ClutchplateDude: Absolutely.

src/Mount.cpp
ClutchplateDude: Yup.

src/Mount.cpp
ClutchplateDude: I'm not sure how we can do that. Certain things need to take place at certain places in the code that needs to be there in either case.

andre-stefanov (approved):

PR #123, opened 2021-07-18
Fix southern hemisphere reversed DEC free slew directions

trishmapow:
North/south directions were reversed. Tested with LCD buttons, OATControl, APT (ASCOM).

### DIFF
--- a/src/Mount.cpp
+++ b/src/Mount.cpp
@@ -2199,7 +2199,7 @@ void Mount::startSlewing(int direction) {
       #endif
 
       if (direction & NORTH) {
-        long targetLocation = sign * 300000;
+        long targetLocation = 300000;
         if (_decUpperLimit != 0) {
           targetLocation = _decUpperLimit;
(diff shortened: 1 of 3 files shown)


src/Mount.cpp
```diff
@@ -2199,7 +2199,7 @@ void Mount::startSlewing(int direction) {
       #endif
 
       if (direction & NORTH) {
-        long targetLocation = sign * 300000;
+        long targetLocation = 300000;
```
ClutchplateDude, line 2202: Hmm, ok. I guess if that's not needed for the Southern hemisphere, then you can delete line 2177.,

ClutchplateDude (changes requested):

src/Mount.cpp
trishmapow: I think it's still needed for RA, so that left/right corresponds to CW/CCW (well this is what I read in docs as expected behaviour).

trishmapow: Forgot to mention, that's in lines 2231 & 2236

ClutchplateDude: Oh boy, i swear i searched for it and found no other uses....

ClutchplateDude (approved):

PR #125, opened 2021-07-23
Fix AZ/ALT/Focus hold currents

CTetford:
Change the hold currents from 10% to 100%.

### DIFF
--- a/src/Mount.cpp
+++ b/src/Mount.cpp
@@ -2438,7 +2438,9 @@ void Mount::loop() {
       // either been stopped, or reached the target.
       _focuserMode = FOCUS_IDLE;
       _focuserWasRunning = false;
+    #if FOCUSER_ALWAYS_ON != 0
       disableFocusMotor();
+    #endif
     }
 
   #endif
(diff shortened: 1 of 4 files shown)


src/Mount.cpp
```diff
@@ -573,7 +573,7 @@ bool Mount::connectToDriver( TMC2209Stepper* driver, const char *driverKind ) {
         _driverAZ->I_scale_analog(false);
     #endif
     LOGV2(DEBUG_STEPPERS, F("Mount: Requested AZ motor rms_current: %d mA"), rmscurrent);
-    _driverAZ->rms_current(rmscurrent, 0.1f); //holdMultiplier = 1 to set ihold = irun
+    _driverAZ->rms_current(rmscurrent, 1.0f); //holdMultiplier = 1 to set ihold = irun
```
ClutchplateDude, line 576: Why do we need 100% holding current? That seems crazy. The focuser has pretty much no load on it.

ClutchplateDude (changes requested):

src/Mount.cpp
CTetford: The default value in the TMCStepper library is 50%. This just keeps consistent with the other motors on the OAT.

We can add yet another config parameter for the user to specify, but I don't see the point.

The only reason this feature will ever be enabled is if the user doesn't want any possibility of the focuser to shift, so logically you'd want it at 100%.

ClutchplateDude (approved):

PR #129, opened 2021-08-02
Az changes

ClutchplateDude:


test/hallread/hallread.ino
```diff
@@ -0,0 +1,11 @@
...
+  // put your setup code here, to run once:
+  pinMode(40, INPUT);
+  Serial.begin(57600);
+}
+
+void loop() {
+  // put your main code here, to run repeatedly:
+  
+  Serial.println(digitalRead(40));
+}
```
andre-stefanov, line 11: I recommend creating a menu for that which displays 1 or 0 for the sensor like Prusa does for e.g. filament sensor. This way a user does not have to reflash some sketches in order to test the sensor. Also:
- we should not use .ino files because our default IDE is platformio / vsc
- `test` folder is a specially treated folder for [pio unit testing](https://docs.platformio.org/en/latest/plus/unit-testing.html), we should not use it for some utility scripts/sketches.

src/b_setup.hpp
```diff
@@ -238,6 +238,10 @@ void setup() {
   #endif
 // end microstepping -------------------
 
+  #if USE_HALL_SENSOR_RA_AUTOHOME == 1
+  pinMode(RA_HOMING_SENSOR_PIN, INPUT);
```
andre-stefanov, line 242: is here indentation missing? at least it does not seem to match indentation of the serial line below.

src/Mount.hpp
```diff
@@ -518,6 +529,14 @@ class Mount {
     #endif
   #endif
 
+#if USE_HALL_SENSOR_RA_AUTOHOME == 1
+  HomingState _homingState;
+  int _homingPinState;
+  int _lastHomingPinState;
+  long _homingPosition[2];
+  long _homingOffsetRA;
```
andre-stefanov, line 537: this looks like it would be perfect for a struct

src/Mount.hpp
```diff
@@ -28,6 +28,13 @@ class TMC2209Stepper;
 #define TARGET_STRING      B01000
 #define CURRENT_STRING     B10000
 
+enum HomingState {
+  HOMING_PIN_IDLE = -1,
+  HOMING_PIN_FINDING_START = 0,
+  HOMING_PIN_FINDING_END = 1,
+  HOMING_PIN_FOUND
```
andre-stefanov, line 35: if we already define explicit values for some of the enums, we should make it consistend and also define one for HOMING_PIN_FOUND. Also there is no reason to start here with -1. In case one would want to use these enum values in boolean expressions, HOMING_PIN_FINDING_START would be the one evaluating to `false` ... i would rather say that IDLE is the one matching `false`.

In general i would not recommend using explicit values here because the enum values should be used by name instead of raw values.

PR #134, opened 2021-09-02
ALT/AZ Fixes for AutoPA v2

CTetford:


Configuration_adv.hpp
```diff
@@ -307,63 +307,68 @@
...
 
-        #define AZ_AUDIO_FEEDBACK 0
-
         #define AZ_STALL_VALUE 10  // adjust this value if the RA autohoming sequence often false triggers, or triggers too late
 
         #ifndef USE_VREF
             #define USE_VREF                                                                                                               \
                 0  //By default Vref is ignored when using UART to specify rms current. Only enable if you know what you are doing.
         #endif
+        #ifndef AZ_ALWAYS_ON
```
ClutchplateDude, line 369: I just did  something similar for the focuser, but wanted the hold current to be a lot  less than the run current. Might it be worth it to use a lower current for holding?

CTetford: What I noticed during testing NEMAs with the alt/az system is that because the movements are so non-continuous (i.e. short pulses from each MAZ/MAL command), unless the steppers stay energized they can't actually make small movements because there isn't enough pressure build up from the motor attempting to turn before it de-energizes again.

In this particular situation I think having the hold current the same as the run current makes sense.

ClutchplateDude (approved):

PR #135, opened 2021-09-04
Focuser changes

ClutchplateDude:
- Focuser can be set to be always on with non-100% holding current.
- Focuser got its own debug channel.
- Added some more logging.


src/b_setup.hpp
```diff
@@ -421,7 +423,8 @@ void setup()
     mount.configureFocusStepper(FOCUSmotorPin1, FOCUSmotorPin2, FOCUS_STEPPER_SPEED, FOCUS_STEPPER_ACCELERATION);
     #endif
     #if FOCUS_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART
-    LOGV1(DEBUG_ANY, F("Configure Focus driver..."));
+    LOGV1(DEBUG_ANY, F("setup(): Configure Focus driver..."));
```
julianneswinoga, line 426: Should this still be `DEBUG_ANY`?

src/Mount.cpp
```diff
@@ -772,8 +772,8 @@ void Mount::configureFocusDriver(Stream *serial, float rsense, byte driveraddres
         #if USE_VREF == 0  //By default, Vref is ignored when using UART to specify rms current.
     _driverFocus->I_scale_analog(false);
         #endif
-    LOGV2(DEBUG_STEPPERS, F("Mount: Requested Focus motor rms_current: %d mA"), rmscurrent);
-    _driverFocus->rms_current(rmscurrent, 1.0f);  //holdMultiplier = 1 to set ihold = irun
+    LOGV2(DEBUG_STEPPERS | DEBUG_FOCUS, F("Mount: Requested Focus motor rms_current: %d mA"), rmscurrent);
+    _driverFocus->rms_current(rmscurrent, FOCUSER_MOTOR_HOLD_SETTING / 100f);  //holdMultiplier = 1 to set ihold = irun
```
julianneswinoga: `100f` is an invalid specifier? Does this code get checked by `matrix_build.py`? If not we need to add it

src/Mount.cpp
```diff
@@ -1663,9 +1671,9 @@ void Mount::setSpeed(StepperAxis which, float speedDegsPerSec)
         }
     #elif AZ_DRIVER_TYPE == DRIVER_TYPE_A4988_GENERIC || AZ_DRIVER_TYPE == DRIVER_TYPE_TMC2209_STANDALONE                                  \
         || AZ_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART
-        float stepsPerSec = speedDegsPerSec * _stepsPerAZDegree;  // deg/sec * u-steps/deg = u-steps/sec
-        LOGV3(DEBUG_STEPPERS, F("STEP-setSpeed: Set AZ speed %f degs/s, which is %f steps/s"), speedDegsPerSec, stepsPerSec);
-        _stepperAZ->setSpeed(stepsPerSec);
+            float stepsPerSec = speedDegsPerSec * _stepsPerAZDegree;  // deg/sec * u-steps/deg = u-steps/sec
```
julianneswinoga: Why the indents here? Was it required for `clang-format`? Seems weird

julianneswinoga (changes requested):
Looks mostly good, only thing I'm really concerned about is the `100f` thing. Also make sure to squash before merging so we don't add 25 commits from this PR

src/Mount.cpp
CTetford: The rms_current input is supposed to be a float value between 0-1. I assume this should be 100.0f rather than 100f?

src/Mount.cpp
ClutchplateDude: Yeah, didn't pass the clang test otherwise

src/Mount.cpp
ClutchplateDude: Well, well, well, TIL. The C spec does require a . for floating point constants.... I guess some compilers are a little more lenient.

src/Mount.cpp
ClutchplateDude: Let me try again.

src/b_setup.hpp
ClutchplateDude: Yes, all the configuration logs for this line are DEBUG_ANY

src/Mount.cpp
ClutchplateDude: Looks like it took.... so extra indentation it doesn't seem to check...

src/Mount.cpp
julianneswinoga: Which compiler were you using?? On my machine gcc fails with an error.
```
~/.platformio/packages/toolchain-atmelavr/bin/avr-g++ --version
avr-g++ (GCC) 7.3.0
```
Could you apply this patch to make `matrix_build.py` test focus builds?
[add_focus_to_matrix_build.txt](https://github.com/OpenAstroTech/OpenAstroTracker-Firmware/files/7118501/add_focus_to_matrix_build.txt)
(`git apply add_focus_to_matrix_build.txt`)
@andre-stefanov can you check that I modified the build script correctly? I'm not 100% sure it's correct

ClutchplateDude: avr-g++ (AVR_8_bit_GNU_Toolchain_3.6.2_1759) 5.4.0

julianneswinoga (changes requested):
Please add build coverage to focuser changes as described in my comment: https://github.com/OpenAstroTech/OpenAstroTracker-Firmware/pull/135#discussion_r703134880

julianneswinoga (approved):

PR #139, opened 2021-10-09
Removing bluetooth and run steppers in main loop

jmoyerman:
# Removing Bluetooth:
Bluetooth hasn't been super popular in the community. I honestly can't find anyone that uses it, so there's no point maintaining code that just doesn't get exercised.

# Removing "Run Steppers In Main Loop":
This feature needs to go away. You just can't run any mount on a processor that doesn't support at least one interrupt for running the stepper loop. The Arduino `loop()` function executes too slow, and much too jittery to be able to track reliably.


julianneswinoga (approved):
No users use these features, we created a deprecation announcement on 10/10/2021: https://discord.com/channels/693401540744183859/715249253081022475/896766645400399892
After a week no users have reached out saying otherwise.

PR #141, opened 2021-10-16
V1.9.34 Changes

ClutchplateDude:
- Added two Meade commands: :XGDP# and :XSDPnnn# to retrieve and set the DEC parking offset.
- Fixed a bug that incorrectly returned a Homing status when the Hall sensor was enabled.
-


julianneswinoga (dismissed):

jmoyerman (approved):
Seems reasonable.

PR #146, opened 2021-11-23
Dual dew heater

CTetford:
Enable HE1 on MKS boards when dew heater is enabled

### DIFF
--- a/boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
+++ b/boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
@@ -182,8 +182,8 @@
 
 //Pin to turn on dew heater MOSFET
 #ifndef DEW_HEATER_PIN1
-    #define DEW_HEATER_PIN1 10
+    #define DEW_HEATER_1_PIN 10
 #endif
 #ifndef DEW_HEATER_PIN2
-    #define DEW_HEATER_PIN2 7
+    #define DEW_HEATER_2_PIN 7
 #endif
--- a/boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
+++ b/boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
@@ -181,6 +181,9 @@
 #endif
 
 //Pin to turn on dew heater MOSFET
-#ifndef DEW_HEATER_PIN
-    #define DEW_HEATER_PIN 10
+#ifndef DEW_HEATER_PIN1
+    #define DEW_HEATER_PIN1 10
+#endif
+#ifndef DEW_HEATER_PIN2
+    #define DEW_HEATER_PIN2 7
 #endif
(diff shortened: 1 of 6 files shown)


boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
```diff
@@ -181,6 +181,9 @@
 #endif
 
 //Pin to turn on dew heater MOSFET
-#ifndef DEW_HEATER_PIN
-    #define DEW_HEATER_PIN 10
+#ifndef DEW_HEATER_PIN1
+    #define DEW_HEATER_PIN1 10
```
andre-stefanov: please rename to DEW_HEATER_1_PIN to keep naming consistend with other definitions.

boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
```diff
@@ -181,6 +181,9 @@
...
 #endif
+#ifndef DEW_HEATER_PIN2
+    #define DEW_HEATER_PIN2 7
```
andre-stefanov: please rename to DEW_HEATER_2_PIN to keep naming consistend with other definitions.

andre-stefanov (changes requested):

boards/AVR_MKS_GEN_L_V2/pins_MKS_GEN_L_V2.h
```diff
@@ -236,6 +236,9 @@
 #endif
 
 //Pin to turn on dew heater MOSFET
-#ifndef DEW_HEATER_PIN
-    #define DEW_HEATER_PIN 10
+#ifndef DEW_HEATER_PIN1
```
ClutchplateDude, line 239: Should we promote DEW_HEATER_PIN to DEW_HEATER_1_PIN? For backwards compat...

src/b_setup.hpp
```diff
@@ -106,7 +106,8 @@ void setup()
 
 //Turn on dew heater
 #if DEW_HEATER == 1
-    digitalWrite(DEW_HEATER_PIN, HIGH);
+    digitalWrite(DEW_HEATER_PIN1, HIGH);
+    digitalWrite(DEW_HEATER_PIN2, HIGH);
```
jmoyerman: While currently harmless, should we support one _or_ two dew heaters?

CTetford: For now I just wanted to get it enabled so it can be used like the other one. In a future update we can add control of each one individually with the option to be on/off by default.

boards/AVR_MKS_GEN_L_V2/pins_MKS_GEN_L_V2.h
CTetford: These names are only used within b_setup and the pins files. If someone has a custom pins file for a different atmega2560 board other than the MKS **and** they're using dew heaters, they'd have to update their pins, but that would be very unlikely.

boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
CTetford: Done

boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
CTetford: Done

ClutchplateDude (approved):

PR #149, opened 2021-12-08
ULN2003 Removal

CTetford:
- Removal of ULN2003 driver support. When using 28BYJ steppers, they must be "bipolar modded" to be compatible with Polulu-style or TMC-style drivers.
- Consolidation of stepper types. The differences in stepper characteristics (SPR, current, and backlash) will be set via config.
- New configuration items are needed: BACKLASH_STEPS, AZ_MICROSTEPPING, and ALT_MICROSTEPPING
- Default motor/driver config is changed to 0.9 deg NEMA SPR and TMC2209 UART, since that's what the kits are.
- Removed a redundant config validation since all motors now work with all drivers


ClutchplateDude (approved):
Looks great! Deployed to my OAT with MKS21 and works fine.

tech-forty-two:
Are you aware that since this change, the Online Configuration Tool generates a configuration file which will no longer compile when choosing 28BYJ-48 motors ?

PR #151, opened 2021-12-29
Add initial unit test framework, MappedDict tests

julianneswinoga:
Platformio uses the Unity test framework:
- http://www.throwtheswitch.org/unity
- https://github.com/ThrowTheSwitch/Unity

I've added some basic tests to the MappedDict utility as a reference and starting point.
Command to run unit tests locally: `pio test -e native -v`. I've also added unit tests to the GitHub workflow so they run automatically.


julianneswinoga:
I had to move `MappedDict.hpp` from  `src/` to `src/libs/MappedDict/`, it's the simplest option to how platformio sets up the unit test environment: https://docs.platformio.org/en/latest/plus/unit-testing.html#shared-code
i.e. we don't want to compile the entire firmware to test a simple header file, and the firmware won't compile under the native architecture.
That means for future unit tests they have to be segregated into their own `libs/` module :woman_shrugging:

.github/workflows/code_format.yml
```diff
@@ -17,8 +17,8 @@ jobs:
           repository: ${{ github.event.pull_request.head.repo.full_name }}
       - uses: DoozyX/clang-format-lint-action@v0.12
         with:
-          source: '. ./src'
-          exclude: './src/libs ./test ./scripts'
+          source: '. ./src ./src/libs'
+          exclude: './src/libs/TimerInterrupt ./test ./scripts'
```
julianneswinoga: Since I changed this file the GitHub workflow will use the unchanged version for this PR but future PRs will have the change incorporated. (this is due to their PR security model)

unit_tests/test_MappedDict.cpp
```diff
@@ -0,0 +1,95 @@
...
+    UNITY_BEGIN();
+    RUN_TEST(test_function_mapped_dict_simple_lookups);
+    RUN_TEST(test_function_mapped_dict_bounds);
+    RUN_TEST(test_function_mapped_dict_zero_size);
+    RUN_TEST(test_function_mapped_dict_pointer_types);
+    UNITY_END();
+}
+
+#if defined(ARDUINO)
+    #include <Arduino.h>
```
julianneswinoga: I haven't tested the unit tests on an embedded platform (currently traveling) but apparently it should just work? If someone wants to test it that'd be great, or we can wait a couple weeks until I can test it.

.github/workflows/code_format.yml
```diff
@@ -47,7 +47,7 @@ jobs:
                1. Run `git apply clang-format-diff.patch` to make the changes
                1. Commit and push up the formatted code
             1. Run `clang-format` locally
-               1. Run the command `bash -c 'shopt -s nullglob; for i in ./{.,src,src/inc,boards/*}/*.{c,cpp,h,hpp}; do clang-format -i $i; done'`
+               1. Run the command `bash -c 'shopt -s nullglob globstar;GLOBIGNORE=./src/libs/TimerInterrupt/*; for i in ./{.,src/**,unit_tests,boards/**}/*.{c,cpp,h,hpp}; do clang-format -i $i; done'`
```
julianneswinoga, line 50: I'm going to make a future PR to make a custom `pio run -t code-format` command that will handle this in a python script, because this is just getting unwieldy

platformio.ini
```diff
@@ -3,6 +3,7 @@
 include_dir = .
 src_dir = ./src
 lib_dir = ./src/libs
+test_dir = ./unit_tests
```
andre-stefanov, line 6: why not just use default "test" dir?

platformio.ini
```diff
@@ -14,7 +15,7 @@ lib_deps =
 	lincomatic/LiquidTWI2@^1.2.7
 	olikraus/U8g2@^2.28.8
 
-[env]
+[common_embedded]
```
andre-stefanov: actually [env] is the default build environment in pio. this is also why we inherited other environments from it (e.g. [env:mega2560]. Now other envs still inherit from the default one but we have no definition for it anymore. I would rather expect following inheritance: 
```
env
   |- embedded
      |- mega2560
      |- ...
   |- native
```

andre-stefanov: embedded env should have:
`test_ignore = test_desktop` so it does not execute tests meant only for the host machine

unit_tests/test_MappedDict.cpp
```diff
@@ -0,0 +1,95 @@
...
+{
+    UNITY_BEGIN();
+    RUN_TEST(test_function_mapped_dict_simple_lookups);
+    RUN_TEST(test_function_mapped_dict_bounds);
+    RUN_TEST(test_function_mapped_dict_zero_size);
+    RUN_TEST(test_function_mapped_dict_pointer_types);
+    UNITY_END();
+}
+
+#if defined(ARDUINO)
```
andre-stefanov: the tests should be separated in folders:
- test_common
- test_embedded
- test_desktop

then appropriate environments should ignore embedded or desktop tests. then only embedded tests should include and rely on Arduino.h

andre-stefanov (changes requested):

platformio.ini
julianneswinoga: Because we had talked about the possibility of having integration tests, and I wanted to set a precedent that this dir will only have unit test in it.

platformio.ini
julianneswinoga: Yeah I figured that almost all of the parameters under `common_embedded` were specific to embedded builds. I guess I could move some of the build flags to a common `env`, since we'd want the native tests to be built the same way

andre-stefanov (dismissed):

anjok:
While you're at it, can you please consider moving DayTime, Longitude and Sidereal to libs/ too so we can do parsing tests natively? And maybe it would better make a "LibOAT" instead as I don't know if you really want to make a directory for each moved class. In particular since Longitude inherits from DayTime.

anjok:
i just saw it uses "String" from Ardiuno... maybe add a copy of WString.h/cpp in tests/ and include conditionally?

anjok:
It would make writing tests just so much faster when you don't have to upload each time...

anjok:
Yeah, with a bit of more targeted includes this would work:

```
//#include "inc/Globals.hpp" // we don't need globals in util classes
#include "../Configuration.hpp" // we probably shouldn't require this in util classes and it's only needed because of debug (?)
#include "Utility.hpp" // this is should be split in math+debug stuff and should include config.h itself of needed
#include "DayTime.hpp"
#include <WString.h> // just load String, not ardiuno
#include <stdlib.h> // strcat etc
#include <math.h> // floating point
```

andre-stefanov (approved):

julianneswinoga:
@anjok
1. It's best practice to keep PRs small and with only one major change (or a couple minor changes bundled together). That way they are easily reviewable and there is less change for conflicts when rebasing/merging. We appreciate the suggestions, perhaps we can incorporate them into a future change!
2. Native tests are really only useful for testing general code, most/all tests should be run on HW to ensure that we get the correct functionality with the HW that we are targeting. As you have seen 8 bit AVR can sometimes differ quite significantly from x86_64!
3. If possible, could you try to avoid having multiple comments in quick succession to another? It's much preferable to have each comment represent a single idea/suggestion rather than stream-of-conscious comments. Thanks :smile:

PR #159, opened 2022-01-19
Disclaimer of auto generation of Meade Command page

christian-kardach:


ClutchplateDude:
We really should not need to update the version and changelog of the firmware for this python script...

PR #161, opened 2022-01-19
Log cleanup

ClutchplateDude:
- Cleaned up logging somewhat
- Ensured all log messages have a category start word.
- Ensured all log messages are stored in Flash memory.

### DIFF
--- a/src/WifiControl.cpp
+++ b/src/WifiControl.cpp
@@ -14,7 +14,7 @@ WifiControl::WifiControl(Mount *mount, LcdMenu *lcdMenu)
 
 void WifiControl::setup()
 {
-    LOGV2(DEBUG_WIFI, F("WIFI: Starting up Wifi As Mode %d\n"), WIFI_MODE);
+    LOGV2(DEBUG_WIFI, F("[WIFI]: Starting up Wifi As Mode %d\n"), WIFI_MODE);
 
     _cmdProcessor = MeadeCommandProcessor::instance();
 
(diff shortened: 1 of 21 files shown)


src/WifiControl.cpp
```diff
@@ -174,31 +174,31 @@ void WifiControl::tcpLoop()
     {
         while (client.available())
         {
-            LOGV2(DEBUG_WIFI, F("WifiTCP: Available bytes %d. Peeking."), client.available());
+            LOGV2(DEBUG_WIFI, F("WIFITCP: Available bytes %d. Peeking."), client.available());
```
andre-stefanov: should not this be `WIFI: TCP:` or `[WIFI][TCP]` ?

src/WifiControl.cpp
```diff
@@ -222,14 +222,14 @@ void WifiControl::udpLoop()
         reply += "@";
         reply += WiFi.localIP().toString();
         LOGV4(DEBUG_WIFI,
-              F("WifiUDP: Received %d bytes from %s, port %d"),
+              F("WIFIUDP: Received %d bytes from %s, port %d"),
```
andre-stefanov, line 225: same as with tcp

andre-stefanov (changes requested):
Maybe we could place tags into [...] ?

cameronswinoga:
Could you rename your commit message to have the same version number as the actual version? (`2.10.4` vs `1.10.4`)

julianneswinoga (approved):
Could you rename your commit message to have the same version number as the actual version? (`2.10.4` vs `1.10.4`)

PR #164, opened 2022-02-03
Fix Enable Pin Logic

### LINKED ISSUE #155: Enable pin incorrect when using generic driver
The firmware doesn't properly enable A4988 drivers when using the generic driver option. Using TMC2209 standalone works for those drivers.

The enable pins are inverted states between those two options.

CTetford:
Use consistent enable pin logic for all drivers.

Increase maximum current for TMC2209 to 2A in accordance with BigTreeTech's published maximum continuous drive current.

Closes #155

### DIFF
--- a/ConfigurationValidation.hpp
+++ b/ConfigurationValidation.hpp
@@ -389,3 +389,31 @@
         #endif
     #endif
 #endif
+
+#if (FOCUS_STEPPER_TYPE != STEPPER_TYPE_NONE)
+    #if !defined(FOCUS_MICROSTEPPING)
+        #error "Altitude microstepping must be defined. Default is 4.0f for 28BYJ steppers, 4.0f for NEMA."
+    #endif
+    #if (FOCUS_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
+        #if defined(FOCUS_MOTOR_CURRENT_RATING)
+            #if (FOCUS_MOTOR_CURRENT_RATING > 2000)
+                #error "The TMC2209 driver is only rated up to 2A output. Delete this error if you know what youre doing"
+            #endif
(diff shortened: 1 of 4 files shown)


ConfigurationValidation.hpp
```diff
@@ -290,8 +290,8 @@
 
 #if (DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART)
```
ClutchplateDude, line 291: What about the Focuser motor?

ClutchplateDude (changes requested):

ConfigurationValidation.hpp
CTetford: Ah my mistake. I forgot about that one. I'll update it.

ConfigurationValidation.hpp
```diff
@@ -388,4 +388,32 @@
                 "ALT_OPERATING_CURRENT_SETTING is not defined. Please define the operating percentage of your motor in you local configuration file using the ALT_OPERATING_CURRENT_SETTING keyword."
         #endif
     #endif
-#endif
\ No newline at end of file
+#endif
+
+#if (FOCUS_STEPPER_TYPE != STEPPER_TYPE_NONE)
+    #if !defined(FOCUS_MICROSTEPPING)
+        #error "Altitude microstepping must be defined. Default is 4.0f for 28BYJ steppers, 4.0f for NEMA."
```
ClutchplateDude: Typo: Altitude

ClutchplateDude (changes requested):

ConfigurationValidation.hpp
CTetford: Fixed

ClutchplateDude (approved):
Looks good.

PR #165, opened 2022-02-03
Fix RA Parking/Homing and Slew Limits

### LINKED ISSUE #152: RA slews off end of ring
calculateRAandDECSteppers appears to not calculate the correct direction to turn the RA axis when attempting to park after a long session of tracking (6+ hrs) 

Exact duration of tracking required to cause this has not been narrowed down yet, but it is repeatable. 

Edit: Confirmed this occurs when attempting to park after 12 hrs of tracking. I'll try to narrow down the window it occurs.

ClutchplateDude:
- Switched GoHome functionality to use stepper positions instead of celestial coordinates.
- Fixed a bug in RA Autohoming
- Cleaned up logging some more


src/Mount.cpp
```diff
@@ -1326,6 +1326,75 @@ void Mount::startSlewingToTarget()
...
+/////////////////////////////////
+//
+// startSlewingToHome
+//
+/////////////////////////////////
+// Calculates movement parameters and program steppers to move to home position.
+// Takes any sync operations that have happened and tracking into account.
+void Mount::startSlewingToHome()
+{
+    if (isGuiding())
```
julianneswinoga: Can we fold this conditional into `stopGuiding()`? Since you'd always use them together

src/Mount.cpp
```diff
@@ -1610,9 +1678,7 @@ void Mount::park()
 /////////////////////////////////
 void Mount::goHome()
 {
-    stopGuiding();
-    setTargetToHome();
```
julianneswinoga, line 1614: We can remove `setTargetToHome` now 😄

src/Mount.cpp
```diff
@@ -2366,6 +2432,38 @@ long Mount::getHomingOffset(StepperAxis axis)
 }
 
 #if USE_HALL_SENSOR_RA_AUTOHOME == 1
+String Mount::getHomingState(HomingState state) const
```
julianneswinoga, line 2436: Could replace with
```C++
String Mount::getHomingState(HomingState state) const
{
    MappedDict<HomingState, String>::DictEntry_t lookupTable[] = {
        {HOMING_MOVE_OFF, F("MOVE_OFF")},
        {HOMING_MOVING_OFF, F("MOVING_OFF")},
        {HOMING_STOP_AT_TIME, F("STOP_AT_TIME")},
        {HOMING_WAIT_FOR_STOP, F("WAIT_FOR_STOP")},
        {HOMING_START_FIND_START, F("START_FIND_START")},
        {HOMING_FINDING_START, F("FINDING_START")},
        {HOMING_FINDING_START_REVERSE, F("FINDING_START_REVERSE")},
        {HOMING_FINDING_END, F("FINDING_END")},
        {HOMING_RANGE_FOUND, F("RANGE_FOUND")},
        {HOMING_FAILED, F("FAILED")},
        {HOMING_SUCCESSFUL, F("SUCCESSFUL")},
        {HOMING_NOT_ACTIVE, F("NOT_ACTIVE")},
    };
    auto strLookup = MappedDict<HomingState, String>(lookupTable, ARRAY_SIZE(lookupTable));
    String rtnStr;
    if (strLookup.tryGet(state, &rtnStr))
    {
        return rtnStr;
    }
    return F("WTF_STATE");
}
```
It's a bit more terse for lookups like this

julianneswinoga (changes requested):

src/Mount.cpp
ClutchplateDude: I like it.

src/Mount.cpp
ClutchplateDude: Well, we can remove `goHome()`

src/Mount.cpp
ClutchplateDude: Yup

src/Mount.cpp
```diff
@@ -1402,7 +1397,10 @@ void Mount::startSlewingToHome()
 /////////////////////////////////
 void Mount::stopGuiding()
 {
-    stopGuiding(true, true);
+    if (isGuiding())
```
julianneswinoga: The conditional should be in `stopGuiding(bool ra, bool dec)`, not `stopGuiding()`
If you're concerned about indentation change, the early return pattern works nicely here:
```C++
if (!isGuiding())
{
    return;
}
```

julianneswinoga (changes requested):

julianneswinoga (dismissed):

CTetford:
@ClutchplateDude have you tested the following scenarios: 
- tracking for 12 hrs and attempting to home?
- tracking for 12 hrs and attempting park (I assume this should be effectively the same thing)
- tracking for 12 hrs and slewing to the current target RA value of what the home is at the time of slew (judging using Stellarium), but not by using the "home" button 
- tracking for 12 hrs and slewing back to a new target in the east

We need to be thorough that it actually fixes all the scenarios that have been reported with slewing incorrectly. It looks like this is only a partial fix so far

ClutchplateDude:
I did not test any 12h scenarios. I did a bunch of sync/slew/sync scenarios and some shorter tracking.
What makes you say this is a partial fix? 
Oh, because this only fixes homing after long tracking? Yeah, we still should take your fixes for the target calculation.
I can integrate them in this PR if you like.

CTetford:
It looks like the revision only affects returning home, not slewing overall. I think the root of the problem is still `calculateRAandDECSteppers`

ClutchplateDude:
> It looks like the revision only affects returning home, not slewing overall. 
Correct, we still need your fix like I said

CTetford:
I'll run some tests on this branch during the day this week. This kind of change needs some thorough testing before we merge, given the implications of it goes wrong

ClutchplateDude:
Sounds good. I am also running some, currently with de-energized steppers, but will try with steppers too

julianneswinoga (dismissed):

src/Mount.cpp
```diff
@@ -1179,19 +1191,20 @@ const DayTime Mount::currentRA() const
...
     float hourPos              = -_stepperRA->currentPosition() / stepsPerSiderealHour;  // u-steps / u-steps/hr = hr
 
-    LOGV4(DEBUG_MOUNT_VERBOSE,
-          F("MOUNT: CurrentRA: Steps/h    : %s (%f x %s)"),
-          String(stepsPerSiderealHour, 2).c_str(),
-          _stepsPerRADegree,
-          String(siderealDegreesInHour, 5).c_str());
-    LOGV2(DEBUG_MOUNT_VERBOSE, F("[MOUNT]: CurrentRA: RA Steps   : %d"), _stepperRA->currentPosition());
-    LOGV2(DEBUG_MOUNT_VERBOSE, F("[MOUNT]: CurrentRA: POS        : %s"), String(hourPos).c_str());
+    // LOGV4(DEBUG_MOUNT_VERBOSE,
```
julianneswinoga: Remove commented code

src/Mount.cpp
```diff
@@ -1179,19 +1191,20 @@ const DayTime Mount::currentRA() const
...
+    //       String(siderealDegreesInHour, 5).c_str());
+    // LOGV2(DEBUG_MOUNT_VERBOSE, F("[MOUNT]: CurrentRA: RA Steps   : %d"), _stepperRA->currentPosition());
+    // LOGV2(DEBUG_MOUNT_VERBOSE, F("[MOUNT]: CurrentRA: POS        : %s"), String(hourPos).c_str());
     hourPos += _zeroPosRA.getTotalHours();
     // LOGV2(DEBUG_MOUNT_VERBOSE,F("[MOUNT]: CurrentRA: ZeroPos    : %s"), _zeroPosRA.ToString());
     // LOGV2(DEBUG_MOUNT_VERBOSE,F("[MOUNT]: CurrentRA: POS (+zp)  : %s"), DayTime(hourPos).ToString());
 
-    bool flipRA = _stepperDEC->currentPosition() < 0;
-    if (flipRA)
+    float degreePos = _stepperDEC->currentPosition() / _stepsPerDECDegree;
```
julianneswinoga: `const float degreePos = (_stepperDEC->currentPosition() / _stepsPerDECDegree) + _zeroPosDEC;`

src/Mount.cpp
```diff
@@ -1222,6 +1235,8 @@ const Declination Mount::currentDEC() const
     //LOGV2(DEBUG_MOUNT_VERBOSE,F("[MOUNT]: CurrentDEC: DEC Steps  : %d"), _stepperDEC->currentPosition());
     //LOGV2(DEBUG_MOUNT_VERBOSE,F("[MOUNT]: CurrentDEC: POS        : %s"), String(degreePos).c_str());
 
+    degreePos += _zeroPosDEC;
```
julianneswinoga: Same thing here, but you could even refactor the entire function here:
```C++
const Declination Mount::currentDEC() const
{
    // u-steps / u-steps/deg = deg
    const float degreePos = (_stepperDEC->currentPosition() / _stepsPerDECDegree) + _zeroPosDEC;

    if (NORTHERN_HEMISPHERE ? degreePos > 0 : degreePos < 0)
    {
        return -degreePos;
    }
    return degreePos;
}
```

src/Mount.cpp
```diff
@@ -1242,25 +1257,33 @@ const Declination Mount::currentDEC() const
...
     long solutions[6];
-    _targetRA  = ra;
     _targetDEC = dec;
+    _targetRA  = ra;
+
+    // Adjust the home RA position by the delta sync position.
+    float raAdjust = ra.getTotalHours() - currentRA().getTotalHours();
+    while (raAdjust > 12)
+    {
+        raAdjust = raAdjust - 24;
```
julianneswinoga: `raAdjust -= 24.0f;`

src/Mount.cpp
```diff
@@ -1242,25 +1257,33 @@ const Declination Mount::currentDEC() const
...
+    }
+    while (raAdjust < -12)
+    {
+        raAdjust = raAdjust + 24;
```
julianneswinoga: `raAdjust += 24.0f;`

src/Mount.cpp
```diff
@@ -1242,25 +1257,33 @@ const Declination Mount::currentDEC() const
...
+    }
+    _zeroPosRA.addHours(raAdjust);
+
+    // Adjust the home DEC position by the delta between the sync'd target and current position.
+    float decAdjust = dec.getTotalDegrees() - currentDEC().getTotalDegrees();
```
julianneswinoga: `const float decAdjust`

src/Mount.cpp
```diff
@@ -1242,25 +1257,33 @@ const Declination Mount::currentDEC() const
...
+    _zeroPosDEC += decAdjust;
 
     long targetRAPosition, targetDECPosition;
-    LOGV3(DEBUG_MOUNT, F("[MOUNT]: Sync Position to RA: %s and DEC: %s"), _targetRA.ToString(), _targetDEC.ToString());
+    LOGV3(DEBUG_MOUNT, F("[MOUNT]: Sync Position to RA: %s and DEC: %s"), ra.ToString(), _targetDEC.ToString());
```
julianneswinoga: Shouldn't this log be both `ra, dec` or `_targetRA, _targetDEC` and not one of either?

src/Mount.cpp
```diff
@@ -1328,46 +1348,110 @@ void Mount::startSlewingToTarget()
...
+    _currentRAStepperPosition  = _stepperRA->currentPosition();
+
+    // Take any syncs that have happened into account
+    long targetRAPosition  = -_homeOffsetRA;
+    long targetDECPosition = -_homeOffsetDEC;
+    LOGV3(DEBUG_STEPPERS, F("[STEPPERS]: startSlewingToHome: Sync op offsets: RA: %l, DEC: %l"), targetRAPosition, targetDECPosition);
+
+    _slewingToHome = true;
+    // Take tracking into account
+    long trackingOffset = _stepperTRK->currentPosition() * RA_SLEW_MICROSTEPPING / RA_TRACKING_MICROSTEPPING;
```
julianneswinoga: `const long trackingOffset`

src/Mount.cpp
```diff
@@ -1328,46 +1348,110 @@ void Mount::startSlewingToTarget()
...
+    _stepperDEC->setMaxSpeed(_maxDECSpeed);
+    LOGV2(DEBUG_STEPPERS, F("[STEPPERS]: startSlewingToHome: Set RA  to MaxSpeed(%d)"), _maxRASpeed);
+    _stepperRA->setMaxSpeed(_maxRASpeed);
+
+    _currentDECStepperPosition = _stepperDEC->currentPosition();
+    _currentRAStepperPosition  = _stepperRA->currentPosition();
+
+    // Take any syncs that have happened into account
+    long targetRAPosition  = -_homeOffsetRA;
+    long targetDECPosition = -_homeOffsetDEC;
```
julianneswinoga: Technically you can remove this variable (and replace with `-_homeOffsetDEC`) but if the name is useful it should at least be `const`

src/Mount.cpp
```diff
@@ -1328,46 +1348,110 @@ void Mount::startSlewingToTarget()
...
+    LOGV4(DEBUG_STEPPERS,
+          F("[STEPPERS]: startSlewingToHome: Adjusted with tracking distance: %l (adjusted for MS: %l), result: %l"),
+          _stepperTRK->currentPosition(),
+          trackingOffset,
+          targetRAPosition);
+
+    moveSteppersTo(targetRAPosition, targetDECPosition);  // u-steps (in slew mode)
+
+    _mountStatus |= STATUS_SLEWING | STATUS_SLEWING_TO_TARGET;
+    _totalDECMove = 1.0f * _stepperDEC->distanceToGo();
```
julianneswinoga:
```C++
_totalDECMove = static_cast<float>(_stepperDEC->distanceToGo());
_totalRAMove  = static_cast<float>(_stepperRA->distanceToGo());
```
instead of the `1.0f *`

src/Mount.cpp
```diff
@@ -1328,46 +1348,110 @@ void Mount::startSlewingToTarget()
...
+/////////////////////////////////
+//
+// stopGuiding
+//
+/////////////////////////////////
 void Mount::stopGuiding(bool ra, bool dec)
 {
-    // Stop RA guide first, since it's just a speed change back to tracking speed
-    if (ra && (_mountStatus & STATUS_GUIDE_PULSE_RA))
+    if (isGuiding())
```
julianneswinoga: You can use an early return pattern to clean up the scope
```C++
if (!isGuiding())
{
    return;
}
```

src/Mount.cpp
```diff
@@ -3163,41 +3270,69 @@ void Mount::calculateStepperPositions(float raCoord, float decCoord, long &raPos
...
+    while (homeTargetDeltaRA > 12)
+    {
+        homeTargetDeltaRA = homeTargetDeltaRA - 24;
     }
+    while (homeTargetDeltaRA < -12)
+    {
+        homeTargetDeltaRA = homeTargetDeltaRA + 24;
+    }
+
     // Map [0 to 24] range to [-12 to +12] range
```
julianneswinoga, line 3305: Seems like we keep doing a lot of this, maybe a future improvement could abstract this kind of operation :smiley:

src/Mount.cpp
```diff
@@ -3740,3 +3884,33 @@ void Mount::testUART_vactual(TMC2209Stepper *driver, int _speed, int _duration)
...
+
+/////////////////////////////////
+//
+// checkRALimit
+//
+/////////////////////////////////
+void Mount::checkRALimit()
+{
+    // Check tracking limits every 5 seconds
+    if (millis() - _lastTRKCheck > 5000)
```
julianneswinoga: Same early return pattern could be used here

src/Mount.cpp
```diff
@@ -3740,3 +3884,33 @@ void Mount::testUART_vactual(TMC2209Stepper *driver, int _speed, int _duration)
...
+    {
+        float trackedHours       = (_stepperTRK->currentPosition() / _trackingSpeed) / 3600.0F;  // steps / steps/s / 3600 = hours
```
julianneswinoga: `trackedHours`, `homeRA`, and `RALimit` can all be `const` :smile: 
(IMO `const float RALimit` instead of `float const RALimit` is the common pattern)

src/Mount.cpp
```diff
@@ -3740,3 +3884,33 @@ void Mount::testUART_vactual(TMC2209Stepper *driver, int _speed, int _duration)
...
+        float homeRA             = _zeroPosRA.getTotalHours() + trackedHours;
+        float const RALimit      = RA_TRACKING_LIMIT;
+        float homeCurrentDeltaRA = homeRA - currentRA().getTotalHours();
+
+        LOGV2(DEBUG_MOUNT_VERBOSE, F("[MOUNT]: checkRALimit: homeRAdelta: %f"), homeCurrentDeltaRA);
+        while (homeCurrentDeltaRA > 12)
+            homeCurrentDeltaRA = homeCurrentDeltaRA - 24;
```
julianneswinoga: `homeCurrentDeltaRA -= 24.0f` helps to show intent here by making the statement a bit less verbose

test/test_embedded/main.cpp
```diff
@@ -0,0 +1,24 @@
+#include <unity.h>
+
+#include "test_sidereal.h"
+
+// void setup()
```
julianneswinoga: Remove commented code

test/test_embedded/main.cpp
```diff
@@ -0,0 +1,24 @@
...
+// void loop()
+// {
+
+// }
+
+int main(int argc, char **argv)
+{
+    UNITY_BEGIN();
+
+    test::sidereal::run();
```
julianneswinoga: Just curious, are these tests actually running on the target? @andre-stefanov had some trouble running embedded tests

julianneswinoga (changes requested):
Mostly minor things, looks awesome overall!

ConfigurationValidation.hpp
```diff
@@ -282,6 +282,13 @@
     #endif
 #endif
 
+// Debugging
+#if DEBUG_SEPARATE_SERIAL == 1
```
julianneswinoga, line 286: Why do we need `DEBUG_SEPARATE_SERIAL`? `DEBUG_SERIAL_PORT` should be enough

boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
```diff
@@ -137,3 +137,8 @@
 #ifndef DEW_HEATER_PIN2
     #define DEW_HEATER_2_PIN 7
 #endif
+
+//Serial port for external debugging
+#ifndef DEBUG_SERIAL_PORT
+    #define DEBUG_SERIAL_PORT Serial3  //D14 (Y-MIN) - TX3 and D15 (Y-MAX) - RX3
```
julianneswinoga: I think the defaults should still be USB serial, debugging for users would get a lot harder if we asked them to hook up an extra serial line

Configuration_adv.hpp
```diff
@@ -559,6 +559,14 @@
     #define DEBUG_LEVEL (DEBUG_NONE)
 #endif
 
+#ifndef DEBUG_SEPARATE_SERIAL
+    #define DEBUG_SEPARATE_SERIAL 0
+#endif
+
+#ifndef DEBUG_SERIAL_BAUDRATE
+    #define DEBUG_SERIAL_BAUDRATE 57600
```
julianneswinoga: Why `57600`? `115200` is a "standard" mid-speed serial connection

src/Declination.cpp
```diff
@@ -93,20 +102,24 @@ Declination Declination::ParseFromMeade(String const &s)
...
+                                              : -(arcSecondsPerHemisphere / 2) + dt.getTotalSeconds();
+    LOGV4(DEBUG_GENERAL, F("[DECLINATION]: Declination.Parse(%s) -> %s (%l)"), s.c_str(), result.ToString(), result.totalSeconds);
     return result;
 }
 
 Declination Declination::FromSeconds(long seconds)
 {
-    seconds += (NORTHERN_HEMISPHERE ? -(arcSecondsPerHemisphere / 2) : (arcSecondsPerHemisphere / 2));
-    return Declination(1.0 * seconds / 3600.0);
+    if (NORTHERN_HEMISPHERE)
```
julianneswinoga: The float conversion should be done upfront to avoid a precision loss:
```C++
Declination Declination::FromSeconds(long seconds)
{
    const auto secondsFloat                 = static_cast<float>(seconds);
    const auto arcSecondsPerHemisphereFloat = static_cast<float>(arcSecondsPerHemisphere);
#if NORTHERN_HEMISPHERE == 1
    return Declination(((arcSecondsPerHemisphereFloat / 2.0f) - secondsFloat) / 3600.0f);
#else
    return Declination(((-arcSecondsPerHemisphereFloat / 2.0f) + secondsFloat) / 3600.0f);
#endif
}
```

src/Mount.cpp
```diff
@@ -1220,13 +1220,14 @@ const Declination Mount::currentDEC() const
 {
     // u-steps / u-steps/deg = deg
     const float degreePos = (_stepperDEC->currentPosition() / _stepsPerDECDegree) + _zeroPosDEC;  // u-steps / u-steps/deg = deg
+    Declination dec(degreePos);
 
-    if (NORTHERN_HEMISPHERE ? degreePos > 0 : degreePos < 0)
-    {
-        return -degreePos;
-    }
+    // if (NORTHERN_HEMISPHERE ? degreePos > 0 : degreePos < 0)
```
julianneswinoga: Remove commented code

julianneswinoga (changes requested):

src/Declination.cpp
ClutchplateDude: Good catch

ConfigurationValidation.hpp
ClutchplateDude: Removed.

src/Mount.cpp

test/test_embedded/main.cpp
julianneswinoga: Reminder to remove these files too

boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
CTetford: I've revised the implementation. DEBUG_SEPARATE_SERIAL is still a boolean option, but I've removed some redundant code and the pins files default to the main serial channel for the debug output.

ConfigurationValidation.hpp
CTetford: See my other comment. I've changed the implementation but kept this flag as an option. If someone wants to specify a non-default debug port directly in their config they can, but the DEBUG_SEPARATE_SERIAL flag is still needed.

src/f_serial.hpp
```diff
@@ -43,8 +43,10 @@ void processSerialData()
             if (buffer[0] == 0x06)
             {
                 LOGV1(DEBUG_SERIAL, F("[SERIAL]: Received: ACK request, replying 1"));
-    #if DEBUG_LEVEL == DEBUG_NONE
-                Serial.print('1');
+                // When not debugging, print the result to the serial port .
+                // When debugging, only print the result to Serial if we're on seperate ports.
+    #if (DEBUG_LEVEL == DEBUG_NONE) || (DEBUG_SEPARATE_SERIAL == 1)
+                Serial.print("1");
```
julianneswinoga: The implementation has changed here I think, I'm not too sure what this code is supposed to be doing but you've changed the serial response from `0x31` (`'1'`) to `0x31 0x0` (`'1'` with null termination)

ClutchplateDude: Oops, good catch.

julianneswinoga (approved):

Issue #166, opened 2022-02-07
Wrong Longitude parsing offset.


anjok:
Apparently there's a 180 deg offset considered when parsing the longitude which is not actually sent, at least not from KStars. KStars simply sends "-11.4" when you're at "11.4" which what's expected from the Meade spec.

Please remove that offset so that LST and HA can be calculated correctly.

You might need to fix the ASCOM driver too and tell all users to upgrade or send the version of the driver in a special command and either add or not add the offset.

JakobPetersson:
Fixes for this are included in #181.

PR #168, opened 2022-02-10
meade offset removed conditionally

anjok:
whole thing is explained here: https://github.com/OpenAstroTech/OpenAstroTracker-Firmware/issues/166

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,6 @@
+**V1.10.8 - Updates**
+- Fix Meade parsing from INDI clients.
+
 **V1.10.6 - Updates**
 - Use consistent enable pin logic for all drivers.
 - Increase maximum current for TMC2209 to 2A in accordance with BigTreeTech's published maximum continuous drive current.
--- a/Configuration_adv.hpp
+++ b/Configuration_adv.hpp
@@ -569,3 +569,9 @@
         #error "Debugging not supported on this platform"
     #endif
 #endif
+
+// define INDI_SUPPORT in the local configuration if you want it to work
+// currently this will fix Meade longitude parsing
+#ifndef INDI_SUPPORT
+    #define INDI_SUPPORT 0
+#endif
--- a/src/Longitude.cpp
+++ b/src/Longitude.cpp
@@ -2,6 +2,12 @@
 #include "Utility.hpp"
 #include "Longitude.hpp"
 
+#if INDI_SUPPORT == 1
+    #define LONGITUDE_OFFSET 0L
+#else
+    #define LONGITUDE_OFFSET 180L
+#endif
+
 //////////////////////////////////////////////////////////////////////////////////////
 //
 // -180..180 range, 0 is at the prime meridian (through Greenwich), negative going west, positive going east
@@ -41,7 +47,7 @@ Longitude Longitude::ParseFromMeade(String const &s)
     DayTime dt = DayTime::ParseFromMeade(s);
 
     //from indilib driver:  Meade defines longitude as 0 to 360 WESTWARD (https://github.com/indilib/indi/blob/1b2f462b9c9b0f75629b635d77dc626b9d4b74a3/drivers/telescope/lx200driver.cpp#L1019)
-    result.totalSeconds = 180L * 3600L - dt.getTotalSeconds();
+    result.totalSeconds = LONGITUDE_OFFSET * 3600L - dt.getTotalSeconds();
     result.checkHours();
 
     LOGV4(DEBUG_GENERAL, F("[LONGITUDE]: Parse(%s) -> %s = %ls"), s.c_str(), result.ToString(), result.getTotalSeconds());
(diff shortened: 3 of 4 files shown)


src/Longitude.cpp
```diff
@@ -2,6 +2,17 @@
 #include "Utility.hpp"
 #include "Longitude.hpp"
 
+// define INDI in the local configuration if you want it to work
+#ifndef INDI
```
julianneswinoga: Could you move this default setting to `Configuration_adv.hpp`? I think the `Misc Configuration` makes sense for this

src/Longitude.cpp
```diff
@@ -2,6 +2,17 @@
...
+    #define INDI 0
```
julianneswinoga: I would rename this `#define` to be a little more specific. `INDI_LONGITUDE_HACK` maybe?

julianneswinoga (changes requested):
FYI in order for all the workflow checks to pass you also need to update `Changelog.md` and `Version.h`

src/Longitude.cpp
anjok: Sure

src/Longitude.cpp
anjok: Actually... it's the opposite, at least from my understanding. What it does is preserve the current parsing on Gg12 -> 180 - 12 for users of the current ASCOM driver, when the correct value should be - 12. The parsing assumes that KStars/Meade/ASCOM will have a 180 - x offset added when in fact it does not.

The whole thing is explained here (and I should referenced this in the subject): https://github.com/OpenAstroTech/OpenAstroTracker-Firmware/issues/166

I just named it INDI (which is a lame name, agreed, maybe INDI_SUPPORT) because I have a feeling there might be more stuff coming in to support KStars and I didn't want flags for all of them.

anjok:
I don't get the needed change for changelog and version. Why is this on me to do that? How would I know which version this is going into? 

Normally, you'd merge some PRs to develop, then decide it's time for a release and do the changelog and version in a commit there...

anjok:
Added version stuff, renamed and moved define  and force-pushed.

julianneswinoga:
> I don't get the needed change for changelog and version. Why is this on me to do that? How would I know which version this is going into?
> 
> Normally, you'd merge some PRs to develop, then decide it's time for a release and do the changelog and version in a commit there...

1. It's up to the author how they would like to describe their changes in the changelog. It's not auto-generated from commits (which I'm assuming is what you are thinking about) because commits are directed more at developers, where the changelog is directed more at users.
2. For changes like this the next version is simply a minor version increment. We try to follow https://semver.org/ convention
3. For full releases the changes are thus very easy to create simply by looking at the prose changes in `Changelog.md` :smile:

julianneswinoga:
> Added version stuff, renamed and moved define and force-pushed.

It would be appreciated if you didn't squash+force-push while a PR is under review (i.e. only do that right before the PR is going to be merged). It makes it very hard to follow incremental changes and review them accordingly.

Changelog.md
```diff
@@ -1,3 +1,6 @@
+**V1.10.8 - Updates**
+- Fix Meade parsing from INDI clients.
+
 **V1.10.6 - Updates**
```
julianneswinoga, line 4: It looks like you need to fetch and rebase your changes onto develop, you created your branch quite a while ago:
```bash
# Add an upstream URL
# This URL is for ssh, change it if you use HTTPS
git remote add upstream git@github.com:OpenAstroTech/OpenAstroTracker-Firmware.git
git fetch -p upstream
git rebase -i origin/develop # Interactive rebase
# Fix merge conflicts
git push -f # Force push is needed because the commits have changed
```

src/Longitude.cpp
```diff
@@ -2,6 +2,12 @@
 #include "Utility.hpp"
 #include "Longitude.hpp"
 
+#if INDI_SUPPORT == 1
+    #define LONGITUDE_OFFSET 0L
```
julianneswinoga, line 6: Could you also add in
```C++
#pragma message("Note: INDI offset used, firmware will not function correctly with ASCOM!")
```
? Just so that we can have traceability if a user is having problems.

src/Longitude.cpp
julianneswinoga: We can re-evaluate at a later date if more INDI-specific changes are needed. More specific is better in the current case.

Configuration_adv.hpp
```diff
@@ -569,3 +569,9 @@
         #error "Debugging not supported on this platform"
     #endif
 #endif
+
+// define INDI_SUPPORT in the local configuration if you want it to work
+// currently this will fix Meade longitude parsing
+#ifndef INDI_SUPPORT
```
julianneswinoga, line 575: Please move this to the `Misc Configuration` section of this file

julianneswinoga (changes requested):

JakobPetersson:
@cameronswinoga Should I create a new PR with these changes if i wish to see them merged?

anjok:
You also need to fix the sign of the utc offset in the meadeparser

anjok:
Like so, but you also need to move the flag to Configuration_avd.hpp (or just skip these conditionals alltogether and just fix it, as this is only about windows users not needing to update the ASCOM driver) : 
``` /////////////////////////////////
 void Mount::setLocalUtcOffset(int offset)
 {
+    #if INDI
+    offset = -offset;
+    #endif
     _localUtcOffset = offset;
```

JakobPetersson:
Fixes for this are included in #181.
I suggest we close this and move forward there.

ClutchplateDude:
Fixes made in #181

PR #169, opened 2022-02-11
Improve Guiding Default Config

CTetford:
- Revise default microstep settings when using UART
- Revise default DEC guide pulse settings to match RA


julianneswinoga:
Will this affect current guide tuning? If so we should outline how the guiding parameters should be adjusted when updating

CTetford:
> Will this affect current guide tuning? If so we should outline how the guiding parameters should be adjusted when updating

It will align the RA and DEC guide step sizes when doing PHD2 calibrations. We could have a statement saying to re-calibrate after applying this update, but realistically people should be calibrating each session regardless.

The microstep change only affects the defaults. It won't affect anyone who has it specified in their local config (which I assume is everyone since it's included by the config generator)

julianneswinoga (dismissed):
I think we should get in the habit of notifying people that the defaults have changed (for example, I don't have `_MICROSTEPPING` defined in my local config)

jmoyerman (approved):

PR #170, opened 2022-02-11
LOGVx -> INFO

anjok:
This is basically `LOGV.\(([^,]+), *F\(("[^"]+")\)([^;]+);` -> `INFO($1, $2$3;` with a few hand edits for multiline and special cases.

```
NO debug:
RAM:   [=====     ]  50.4% (used 4132 bytes from 8192 bytes)
Flash: [===       ]  32.5% (used 84892 bytes from 261120 bytes)

INFO debug:
RAM:   [=====     ]  54.1% (used 4430 bytes from 8192 bytes)
Flash: [=====     ]  45.3% (used 118414 bytes from 261120 bytes)

INFO debug, no using F(fmt) (jikes!):
RAM:   [==========]  202.6% (used 16600 bytes from 8192 bytes)
Flash: [=====     ]  45.2% (used 118086 bytes from 261120 bytes)
```

### DIFF
--- a/src/DayTime.cpp
+++ b/src/DayTime.cpp
@@ -17,7 +17,7 @@ DayTime DayTime::ParseFromMeade(String const &s)
     DayTime result;
     int i    = 0;
     long sgn = 1;
-    LOGV2(DEBUG_MEADE, F("[DAYTIME]: Parse Coord from [%s]"), s.c_str());
+    INFO(DEBUG_MEADE, "[DAYTIME]: Parse Coord from [%s]", s.c_str());
     // Check whether we have a sign. This should be able to parse RA and DEC strings (RA never has a sign, and DEC should always have one).
     if ((s[i] == '-') || (s[i] == '+'))
     {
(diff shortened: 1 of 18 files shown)


anjok:
Apparently it works:

[17263]:3535: [SYSTEM]: Boot complete!

I did get warnings and errors with non-MKS, but that's because the config is borked and the format strings are wrong (double<->float etc) and it's not for this patch to fix that.

I also tried to compile with Arduino IDE, but that failed because some libs aren't installed and I don't know whch ones to get and how. Which again is unrelated to this particular patch.

I don't know how well your CI tasks work and if they even enable debug mode, so it might not be all to helpful

julianneswinoga (changes requested):
Thanks for the PR! One request off the bat though:
For large changes with automated tools, ideally what happens is
- one commit with the script in the message, in addition to an explanation of the changes
  - this way anyone can verify the changes by running the script locally and observing that no diff is generated
  - example, [a25945](https://github.com/OpenAstroTech/OpenAstroTracker-Firmware/commit/a25945b9984f8ca2ac582ecc7d634635d7ce9134):
```
Reformat codebase according to .clang-format

Using script:
bash -c 'shopt -s nullglob; for i in ./{.,src,src/inc,boards/*}/*.{c,cpp,h,hpp}; do clang-format -i $i; done'
```
- A following commit with any other changes that you may want in the PR (manual fixes for example)

Could you update your PR to follow this format please? :smiley:

anjok:
This wasn't a script, I just used the regex in the description inside of VSCode. And it took me about an hour which I don't want to do again.

As I said, there was bit of hand-editing because you can't use (foo==bar ? "1" : "2") in the format string and some "FOO:" i edited to "[FOO]:" which is done on a regular basis inline with other commits.

And the clang script spews out a bunch of:

```
-    enum ItemFlag
-    {
+    enum ItemFlag {
```
and nothing else which are totally unrelated to my change.I did already run the patch from the CI system.

anjok:
Which btw is weird as all the other `{` are on their own line?

anjok:
What I can do is amend the commits with the regex and squash them, but I won't be pulling this apart again.

julianneswinoga:
1. I'm sorry but this kind of large stylistic change is not worth reviewing unless it is machine-verifiable. I could try my hand at a script to do the conversion if you'd like, I've got a bit of experience with regular expressions :smile: 
2. The brace wrapping is controlled by our `.clang-format` configuration, to ensure consistent code style

anjok:
I understand your point, but it was 1+ hour of pretty tedious work and I won't be redoing this. 
If you want to do this yourself, sure:
- The regex is listed above, although I don't see what's "machine verifiable" about this. 
- the only real trick is that apparently you can't use `__VA_ARGS__` but have to use `##_VA_ARGS_` instead because the PIO C-compiler is too old.
- IIRC, that time was mostly spent on multiline edits (search for `LOG.\(` and then  also remove the `F(x)` for the format string as that's handled handled in `INFO(...)`)
- but also you can't do`foo ? "BAR" : "BAZ"` as you're doing in some strings so these have to be redone.

anjok:
And the enum thing may be in your config now, but as I said it's unrelated to my change (and it's weird in that it changes ALL current enums). So I don't see why it should be in that patch, it would be better if you did that yourself prior to any further code changes.

anjok:
The `##__VA_ARGS__` thing is from here: https://stackoverflow.com/questions/5588855/standard-alternative-to-gccs-va-args-trick. I didn't reading beyond the question since it apparently works.

anjok:
And this is actually only the first half of changes that would need to be done. 
The second part would be moving `DEBUG_VERBOSE` and `DEBUG_MOUNT_VERBOSE` -> `DEBUG(...)` and probably `DEBUG_ANY` -> `DEBUG_GENERAL`. 
but those are intent changes which should be done by the owner. But this will work for now.

julianneswinoga:
That's alright, it's understandable that you don't want to re-do tedious work. We appreciate the effort though! :smiley:

PR #171, opened 2022-02-12
Change LOGVx to LOG

julianneswinoga:
Taking over the torch from #170

Breakdown of the regex used for the replacement:
## Match portion
```regex
LOGV\d\((?<log_level>.*?),\s*(?:F?\(?(?<fmt_str>\".*?\")\)?|(?<fmt_tern>.*?\?.*?\s:\s*.*?))(?<extra_args>,.*?)??\);
```
* Where the important groups are:
    * `<log_level>` (`$1`), `<fmt_str>` (`$2`), `<fmt_tern>` (`$3`), `<extra_args>` (`$4`)
* Matches `F()` macros, but not in a group and thus are not carried over to the replacement
* Matches ternaries in the string format field, but doesn't replace them. I've fixed those in a manual update
## Replacement portion:
```regex
LOG($1, $2$3$4);
```
* Just uses the groups as described above

----

You can use the following website breakdown to get a deeper understanding of what is happening, as well as the test cases that I used:
* https://regex101.com/r/XMdaQa/1
* https://www.debuggex.com/r/JBylBw2bmj4Uoux4

You can verify the changes by:
```bash
git checkout 8a1741d5860^ # Check out 1 commit before mass replacement
 # Echo commit message to file, cutting out everything before the embedded bash script
git log --format=%B -n 1 8a1741d5860 |tail -n +4 > regex.sh
chmod +x regex.sh && ./regex.sh # Run the script
git diff 8a1741d5860 -- # Show that the current changes are identical to the commit
```

I also added `DEBUG_LEVEL` to `matrix_build.py` to ensure that these kind of changes will be able to be compiled

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -29,7 +29,7 @@ BraceWrapping:
   AfterCaseLabel:  true
   AfterClass:      true
   AfterControlStatement: Always
-  AfterEnum:       false
+  AfterEnum:       true
   AfterFunction:   true
   AfterNamespace:  true
   AfterObjCDeclaration: true
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,6 @@
+**V1.11.3 - Updates**
+- Change logging macro LOGVx to LOG
+
 **V1.11.2 - Updates**
 - Cache build files by default
 
--- a/Version.h
+++ b/Version.h
@@ -3,4 +3,4 @@
 // Also, numbers are interpreted as simple numbers.                        _   __   _
 // So 1.8 is actually 1.08, meaning that 1.12 is a later version than 1.8.  \_(..)_/
 
-#define VERSION "V1.11.2"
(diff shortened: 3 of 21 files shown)


anjok:
I would really go with INFO(...) (and DEBUG(...) which adds DEBUG_VERBOSE to the flags) instead of LOG(...).

ClutchplateDude (dismissed):

julianneswinoga:
Appears to be some weirdness with `clang-format` and BraceWrapping::AfterEnum https://reviews.llvm.org/D93938
My local version:
```console
$ clang-format --version
clang-format version 13.0.1
```
GitHub Actions version:
```
"--clang-format-executable" "/clang-format/clang-format12"
```

ClutchplateDude (approved):
I like it

PR #178, opened 2022-03-20
Task/cs/general cleanup

julianneswinoga:
Two commits, first is non-code changes and the second is removing unused code (as directed by CLion and reviewed by me). Of course let me know if the code is vestigial or has some historical meaning that I don't know about :smiley:


ClutchplateDude:
And this compiles? 😁

julianneswinoga:
> And this compiles? grin

Yep, gotta love the compile checks :smile: 
Most (if not all) of the removals are easily seen to not be used via searches

ClutchplateDude:
I kinda was joking, since we do have the CI pipe. I was surprised that `_currentDECPosition` isn't used anywhere, but `_currentRAPosition` is... but turns out we could eliminate `_currentRAPosition` as well and just use a local variable in `Mount::loop() `instead, since it's only used to correct backlash there.
Looks like there's a version problem somewhere too.

julianneswinoga:
Yep, good points! I haven't updated the version stuff yet since this was kind of exploratory changes

Configuration.hpp
```diff
@@ -10,11 +10,9 @@
 * and not need to worry about a new version from Git overwriting your setup. 
 * There are multiple ways to define a local config file:
 *  - For all boards/hardware configs:
-*    Create a file called Configuration_local.hpp (best to copy configuration_sample_local.hpp and 
-*    change it as needed)
+*    Create a file called Configuration_local.hpp (best to visit https://config.openastrotech.com/)
 *  - Specific to a board:
-*    Create a file called Configuration_local_<board>.hpp, where <board> is either 'mega' or 
-*    'esp32' or (here, too, best to copy Configuration_sample_local.hpp and change it as needed). 
+*    Create a file called Configuration_local_<board>.hpp, where <board> is either 'mega' or 'esp32'
```
ClutchplateDude: Let's mention our recommended board here... I use Configuration_local_mksgenlv21.hpp for my config

ClutchplateDude (changes requested):

LocalConfiguration.hpp
```diff
@@ -6,8 +6,8 @@
 // Don't load local configuration file
 #elif defined(MATRIX_LOCAL_CONFIG)
     #include "Configuration_local_matrix.hpp"
-#elif BOARD == BOARD_AVR_MEGA2560 && __has_include("Configuration_local_mega2560.hpp")
-    #include "Configuration_local_mega2560.hpp"
+#elif BOARD == BOARD_AVR_RAMPS && __has_include("Configuration_local_ramps.hpp")
```
ClutchplateDude, line 9: Why retire the MEGA2560?

julianneswinoga: Since you can't actually build for it any more (@CTetford removed it in #163 ), it was fully replaced with the ramps build

ClutchplateDude (approved):
LGTM

PR #179, opened 2022-03-21
Fix description of Meade command parameters

JakobPetersson:
After using/reading through the Meade command documentation  I have a very small fix.

### DIFF
--- a/src/MeadeCommandProcessor.cpp
+++ b/src/MeadeCommandProcessor.cpp
@@ -375,7 +375,7 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
 //      Returns:
 //        "1Updating Planetary Data#                              #"
 //      Parameters:
-//        "HHMM" is the month
+//        "MM" is the month
 //        "DD" is the day
 //        "YY" is the year since 2000
 //


anjok:
You also need to fix the utc offset

JakobPetersson:
> You also need to fix the utc offset

I am unsure of what related documentation of UTC offset you mean could be fixed at the same time. Can you clarify?

Edit: I understand which change you mean. But that appears to be an incorrect implementation of the protocol, and should be handled as a separate issue. Unrelated to this small fix.

ClutchplateDude:
> > You also need to fix the utc offset
> 
> I am unsure of what related documentation of UTC offset you mean could be fixed at the same time. Can you clarify?
> 
> Edit: I understand which change you mean. But that appears to be an incorrect implementation of the protocol, and should be handled as a separate issue. Unrelated to this small fix.

You don't *NEED* to fix anything, Jakob. Just submit what you feel comfortable with and that helps with the code base.

JakobPetersson:
> You don't _NEED_ to fix anything, Jakob. Just submit what you feel comfortable with and that helps with the code base.

You don't have to worry @ClutchplateDude :) Thank you for checking in on this PR.

Should we keep the PR going?
Merge it?
Or move these changes into #180?

JakobPetersson:
Fixes included in #181, closing.

PR #180, opened 2022-03-21
Protocol INDI corrections

JakobPetersson:
(Updated)
A new take on previous PR: #168

Preface:
* I am not an experienced c/c++ programmer.
* This PR is open to any and all suggestions.

From the sources I have found, the current implementation of UTC Offset protocol is inverted compared to Meade protocol.
This is changed in the first commit.

From the Meade 2002 protocol documents and sources the 0-360 westward format does not appear to be used. It uses a format that is basically the opposite of cartography format.
This is changed in the second commit.

Right now the changes are written as a default  general fix, with an optional fallback to be compatible with older(current) ASCOM driver. This logic could be inverted to be backwards compatible by default instead?  

Questions:
What route does the maintainers want to go with regards of these potential changes to the protocol?
My understanding is that you have tried to implement a protocol that follows the LX200 and that this protocol is implemented here in the firmware as well as the ASCOM driver.
So you do not strictly _have_ to follow the LX200 protocol if you choose not to, but making these changes seems to be a step towards following the LX200 more closely.
However, making these changes would also require changing the ASCOM driver and OATControl in https://github.com/OpenAstroTech/OpenAstroTracker-Desktop.

### DIFF
--- a/src/Longitude.cpp
+++ b/src/Longitude.cpp
@@ -76,16 +76,11 @@ const char *Longitude::ToString() const
 
 const char *Longitude::formatString(char *targetBuffer, const char *format, long *) const
 {
+#if USE_OLD_ASCOM_DRIVER_COMPATIBLE_PROTOCOL == 1
     long secs = totalSeconds;
 
-#if USE_OLD_ASCOM_DRIVER_COMPATIBLE_PROTOCOL == 1
     // from indilib driver:  Meade classic handset defines longitude as 0 to 360 WESTWARD (https://github.com/indilib/indi/blob/1b2f462b9c9b0f75629b635d77dc626b9d4b74a3/drivers/telescope/lx200driver.cpp#L1019)
     secs = 180L * 3600L - secs;
-#else
-    // from indilib driver: Meade API expresses East Longitudes as negative, West Longitudes as positive. (https://github.com/indilib/indi/blob/8beeeb9c5809063929804b5066f5b8ba6696bcd0/drivers/telescope/lx200driver.cpp#L1180)
-    // Source: https://www.meade.com/support/LX200CommandSet.pdf from 2002 at :Gg#
-    secs = -secs;
-#endif
 
     long degs = secs / 3600;
     secs      = secs - degs * 3600;
@@ -93,4 +88,17 @@ const char *Longitude::formatString(char *targetBuffer, const char *format, long
     secs      = secs - mins * 60;
 
     return formatStringImpl(targetBuffer, format, '\0', degs, mins, secs);
+#else
+    // from indilib driver: Meade API expresses East Longitudes as negative, West Longitudes as positive. (https://github.com/indilib/indi/blob/8beeeb9c5809063929804b5066f5b8ba6696bcd0/drivers/telescope/lx200driver.cpp#L1180)
+    // Source: https://www.meade.com/support/LX200CommandSet.pdf from 2002 at :Gg#
+    long secs = -totalSeconds;
+    char sgn  = secs < 0 ? '-' : '+';
+    secs      = labs(secs);
+    long degs = secs / 3600;
+    secs      = secs - degs * 3600;
+    long mins = secs / 60;
+    secs      = secs - mins * 60;
+
+    return formatStringImpl(targetBuffer, format, sgn, degs, mins, secs);
+#endif
 }
--- a/src/Longitude.cpp
+++ b/src/Longitude.cpp
@@ -40,8 +40,16 @@ Longitude Longitude::ParseFromMeade(String const &s)
     // Use the DayTime code to parse it.
     DayTime dt = DayTime::ParseFromMeade(s);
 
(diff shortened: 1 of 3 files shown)


src/Longitude.cpp
```diff
@@ -40,8 +40,16 @@ Longitude Longitude::ParseFromMeade(String const &s)
...
     DayTime dt = DayTime::ParseFromMeade(s);
 
-    //from indilib driver:  Meade defines longitude as 0 to 360 WESTWARD (https://github.com/indilib/indi/blob/1b2f462b9c9b0f75629b635d77dc626b9d4b74a3/drivers/telescope/lx200driver.cpp#L1019)
+#if USE_OLD_ASCOM_DRIVER_COMPATIBLE_PROTOCOL
+    // from indilib driver:  Meade classic handset defines longitude as 0 to 360 WESTWARD (https://github.com/indilib/indi/blob/1b2f462b9c9b0f75629b635d77dc626b9d4b74a3/drivers/telescope/lx200driver.cpp#L1019)
     result.totalSeconds = 180L * 3600L - dt.getTotalSeconds();
+#else
+    // from indilib driver: Meade API expresses East Longitudes as negative, West Longitudes as positive. (https://github.com/indilib/indi/blob/8beeeb9c5809063929804b5066f5b8ba6696bcd0/drivers/telescope/lx200driver.cpp#L1180)
+    // Source: https://www.meade.com/support/LX200CommandSet.pdf from 2002 at :Gg#
+    #pragma message("Note: INDI-compatible longitude format used, firmware will not function correctly with old ASCOM driver!")
```
julianneswinoga: Please move the `#pragma`s to `Configuration_adv.hpp` (so the message only shows once)

src/Longitude.cpp
```diff
@@ -40,8 +40,16 @@ Longitude Longitude::ParseFromMeade(String const &s)
     // Use the DayTime code to parse it.
     DayTime dt = DayTime::ParseFromMeade(s);
 
-    //from indilib driver:  Meade defines longitude as 0 to 360 WESTWARD (https://github.com/indilib/indi/blob/1b2f462b9c9b0f75629b635d77dc626b9d4b74a3/drivers/telescope/lx200driver.cpp#L1019)
+#if USE_OLD_ASCOM_DRIVER_COMPATIBLE_PROTOCOL
```
julianneswinoga: Best practice is to add a full conditional (` == 1`) to the `#if` check

julianneswinoga (changes requested):
First off, thank you for this contribution! I really appreciate the time taken to write a detailed description. The devs are discussing this change internally and we might have a solution that satisfies both INDI and ASCOM without compile-time changes. I don't have much experience in the Meade protocol so I'll leave the functionality decisions to @ClutchplateDude but I have some minor code change requests :smile:

ClutchplateDude:
> The devs are discussing this change internally and we might have a solution that satisfies both INDI and ASCOM.... 

We're not sure that's possible, because of the UTC inversion bug. But it does need some thought around release since it breaks ASCOM and OATControl and we would need to release all three at the same time.

JakobPetersson:
@cameronswinoga I have made the requested changes in a new commit to keep the history in the PR correct. If anything is to be merged later I suggest I rebase it then to make it look nicer.

> We're not sure that's possible, because of the UTC inversion bug. But it does need some thought around release since it breaks ASCOM and OATControl and we would need to release all three at the same time.
I understand this is tricky to do, any API-changes needs the be well understood before committing to them.

As I have understood it, a rework of the firmware is in the very early stages. Perhaps this could be used to shape that work as well. Maybe there is a use for a more formal abstraction layer between the USB/Serial-protocol and the internal functionality of the mount. Allowing us to keep and test the LX200 translations separate from mount-internal-functionality.

ClutchplateDude:
> As I have understood it, a rework of the firmware is in the very early stages. 
Yes, still very early... but we need to maintain the current and keep eliminating bugs....

src/Longitude.cpp
```diff
@@ -71,8 +78,14 @@ const char *Longitude::formatString(char *targetBuffer, const char *format, long
...
     long secs = totalSeconds;
 
-    // Map to 0..360 westwards
+#if USE_OLD_ASCOM_DRIVER_COMPATIBLE_PROTOCOL == 1
+    // from indilib driver:  Meade classic handset defines longitude as 0 to 360 WESTWARD (https://github.com/indilib/indi/blob/1b2f462b9c9b0f75629b635d77dc626b9d4b74a3/drivers/telescope/lx200driver.cpp#L1019)
     secs = 180L * 3600L - secs;
+#else
+    // from indilib driver: Meade API expresses East Longitudes as negative, West Longitudes as positive. (https://github.com/indilib/indi/blob/8beeeb9c5809063929804b5066f5b8ba6696bcd0/drivers/telescope/lx200driver.cpp#L1180)
+    // Source: https://www.meade.com/support/LX200CommandSet.pdf from 2002 at :Gg#
+    secs = -secs;
```
JakobPetersson: This "solution" is broken since formatStringImpl will break on non-positive inputs to degs, mins, secs.

JakobPetersson: "Fixed" in latest commit. I say "fixed" since it is very ugly. I am more looking forward to the changes that @ClutchplateDude are doing in the `meade-fix` branch.

JakobPetersson:
Changes included in #181, closing.

PR #181, opened 2022-03-24
Fixes to Meade Longitude and UC handling

ClutchplateDude:
- Corrected Longitude parsing to account for sign
- Corrected Longitude output to provide sign
- Corrected inverted UTC offset
- Corrected handshake response to 0x06 to be P for Polar mode
- EEPROM debug code caused crash due to using F() macro in args
- Added specific Longitude output function for Meade
- Fixed comment on Meade :SC command

### DIFF
--- a/src/DayTime.cpp
+++ b/src/DayTime.cpp
@@ -267,13 +267,14 @@ const char *DayTime::formatStringImpl(char *targetBuffer, const char *format, ch
         i++;
     }
 
-    if (degs >= 100)
+    long absdegs = labs(degs);
+    if (absdegs >= 100)
     {
-        achDegs[i++] = '0' + min(9L, (degs / 100));
-        degs         = degs % 100;
+        achDegs[i++] = '0' + min(9L, (absdegs / 100));
+        absdegs      = absdegs % 100;
     }
 
-    printTwoDigits(achDegs + i, degs);
+    printTwoDigits(achDegs + i, absdegs);
     printTwoDigits(achMins, mins);
     printTwoDigits(achSecs, secs);
 
--- a/src/DayTime.cpp
+++ b/src/DayTime.cpp
@@ -294,6 +294,11 @@ const char *DayTime::formatStringImpl(char *targetBuffer, const char *format, ch
                     {
                         switch (macro)
                         {
+                            case '+':
+                                {
+                                    *p++ = (degs < 0 ? '-' : '+');
+                                }
+                                break;
                             case 'd':
                                 {
                                     strcpy(p, achDegs);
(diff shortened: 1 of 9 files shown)


src/EPROMStore.cpp
```diff
@@ -121,9 +121,9 @@ void EEPROMStore::displayContents()
     uint16_t marker = readUint16(MAGIC_MARKER_AND_FLAGS_ADDR);
     LOG(DEBUG_EEPROM, "[EEPROM]: Magic Marker: %x", marker);
 
-    LOG(DEBUG_INFO, "[EEPROM]: Values? %s", ((marker & MAGIC_MARKER_MASK) == MAGIC_MARKER_VALUE) ? F("Yes") : F("No"));
-    LOG(DEBUG_INFO, "[EEPROM]: Extended values? %s", ((marker & EXTENDED_FLAG) == EXTENDED_FLAG) ? F("Yes") : F("No"));
-    LOG(DEBUG_INFO, "[EEPROM]: IsPresent(EXTENDED)? %s", (EEPROMStore::isPresent(EXTENDED_FLAG) ? F("Yes") : F("No")));
+    LOG(DEBUG_INFO, "[EEPROM]: Values? %s", ((marker & MAGIC_MARKER_MASK) == MAGIC_MARKER_VALUE) ? "Yes" : "No");
```
julianneswinoga, line 124: Why're you removing `F()`?

src/DayTime.cpp
```diff
@@ -294,6 +295,11 @@ const char *DayTime::formatStringImpl(char *targetBuffer, const char *format, ch
                     {
                         switch (macro)
```
julianneswinoga, line 296: More of a general question, but why are we doing our own string formatting? Is this stuff specific to an interface?

julianneswinoga (changes requested):

src/EPROMStore.cpp
ClutchplateDude: Caused a crash in the arguments list. I don't know why, but after removing it, it no longer crashed.

src/DayTime.cpp
ClutchplateDude: Not sure what you mean? As opposed to using sprintf()?

src/EPROMStore.cpp
julianneswinoga: Huh, good to know I guess :+1:

src/DayTime.cpp
julianneswinoga: Yeah, the standard library of string formatting functions

ClutchplateDude: That was two years ago, I can't remember yesterdays breakfast.... there was a reason in that I couldn't get it to work correctly on the Arduino. I don't remember why, though.

julianneswinoga: Lol fair enough, if it's not interface specific I may look into removing this stuff and changing over to standard string formatting

julianneswinoga (approved):

PR #182, opened 2022-03-25
Change default pin for RA_HOMING_SENSOR_PIN to pin 3 on mksgenlv21

JakobPetersson:
I am proposing to change the default pin for hall sensor from 53 to 3.

The board is designed to use pin 3 for homing the X-axis motor (RA-axis on OAT). And when using a hall sensor the 3-pin JST connector work perfectly in this port without splitting the cable to connect signal, voltage and ground in different places. It is a clean install.

**Premise:**
This change is based on the premise that the most common use-case would be to only use one method of homing per axis (at a time).

My assumption is that you would reasonably only use either a hall sensor for very exact homing, an endstop, or the TMC sensorless homing (If it is implemented).

On mksgenlv21 a jumper controls if the diag-pin is connected to pin 3 or not. Jumper missing, free to use for hall sensor. Jumper in place, connected to X-diag in use for potential sensorless homing with TMC stepper.

**Other:**
This wiki page would need to be updated along with this change:
https://wiki.openastrotech.com/OpenAstroTracker/Addons/RAHallSensor

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,6 @@
+**V1.11.6 - Updates**
+- Change RA Auto Home addon default pin to pin D11 for MKS Gen L v2.1
+
 **V1.11.5 - Updates**
 - Corrected Longitude parsing to account for sign
 - Corrected Longitude output to provide sign
--- a/Version.h
+++ b/Version.h
@@ -3,4 +3,4 @@
 // Also, numbers are interpreted as simple numbers.                        _   __   _
 // So 1.8 is actually 1.08, meaning that 1.12 is a later version than 1.8.  \_(..)_/
 
-#define VERSION "V1.11.5"
+#define VERSION "V1.11.6"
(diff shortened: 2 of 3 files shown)


JakobPetersson:
Rebased on 1.11.5 from develop branch.

CTetford:
We discussed this internally and agree with your reasoning, but each of the X/Y/Z min/max pins have been allocated already for future features (they're shared with the diag pins). 

X-min/max and Z-min/max are going to be used for limit switches for each axis on the OAM.
Y-min/max are already used for the separate debug serial port which is needed for development purposes.

The best alternative is pin D11. It is adjacent to a 5V and gnd pin, so a single 3-pin Dupont connector can be used for the hall sensor.

Please change the pin and version number to the next available increment and we'll get this merged.

JakobPetersson:
I believe I have made the requested changes now. @CTetford

PR #188, opened 2022-06-14
Make steps/deg storage independant of microstep setting.

ClutchplateDude:
- Change storage of RA and DEC steps/degree to be independent of microstepping settings set in config.


src/EPROMStore.cpp
```diff
@@ -383,8 +385,16 @@ float EEPROMStore::getRAStepsPerDegree()
 {
     float raStepsPerDegree(RA_STEPS_PER_DEGREE);  // Default value
 
-    if (isPresent(RA_STEPS_FLAG))
+    // Latest version stores 100x steps/deg for 256 MS
+    if (isPresentExtended(RA_NORM_STEPS_MARKER_FLAG))
+    {
+        float factor     = 25600.0 / RA_TRACKING_MICROSTEPPING;
```
julianneswinoga: Refactor magic number `25600.0` into something meaningful

src/EPROMStore.cpp
```diff
@@ -399,23 +409,32 @@ float EEPROMStore::getRAStepsPerDegree()
...
 // If it is not present then the default uncalibrated DEC_STEPS_PER_DEGREE value is returned.
 float EEPROMStore::getDECStepsPerDegree()
 {
     float decStepsPerDegree(DEC_STEPS_PER_DEGREE);  // Default value
 
-    if (isPresent(DEC_STEPS_FLAG))
+    if (isPresentExtended(DEC_NORM_STEPS_MARKER_FLAG))
+    {
+        // This version stored 100x steps/deg for 256 MS
+        float factor      = 25600.0 / DEC_GUIDE_MICROSTEPPING;
```
julianneswinoga: Did you test this with different `*_MICROSTEPPING` `#define`s? Do you need to clear EEPROM if you change that stuff, cause the math gets out of sync with the EEPROM?

julianneswinoga (changes requested):

src/EPROMStore.cpp
ClutchplateDude: Yes, I flashed my board with different MS settings. The whole point of this PR is to NOT have to reset the EEPROM when changing MS settings....

src/EPROMStore.cpp
ClutchplateDude: Any suggestions? Scale256x100? NormedTo256MSx100? The comment above the line explains what it is. StorageScaling?

src/EPROMStore.cpp
julianneswinoga: Sure! I don't totally understand our EEPROM functionality so I'm just checking that you did your due diligence :smile:

src/EPROMStore.cpp
julianneswinoga: So we're assuming 256 is the maximum supported microstep setting? (I know it's the max on TMCs)
Maybe have a constant in `EPROMStore.hpp` `maxSupportedMicrostepping = 256`, then something like `storageScaling = 100.0`?

julianneswinoga: `const float factor`

ClutchplateDude: It could support higher microstepping. I could also normalize it to 1.0 (instead of 25600) and store a float.

julianneswinoga:
I'm confused, how is b2976a43 related to the review comments I made?

ClutchplateDude:
I screwed up. Fixed.

julianneswinoga (approved):

PR #193, opened 2022-09-29
Improved LCD support and made some park/homing fixes

ClutchplateDude:
- Added RA Autohoming support to Guided startup
- Added DEC Offset homing support to Guided startup
- Defaulted Auto-home questions to YES on LCD during Guided Startup
- Fixed bug that did not honor TrackingOffOnBoot flag when using Guided startup

- Added DEC Offset homing support to CTRL menu
- Removed SetParkingPos from CAL menu
- Added SetDecHomingOffset to CAL menu
- Fixed Park command to actually park DEC after homing
- Fixed a bug that made DEC parking go in the wrong direction
- Made Unpark go to home position after enabling Tracking

- Fixed SetLimit commands to actually use passed limits
- Fixed bug that did not change submenu items in the CTRL menu as button up and down were pressed
- Renamed some menu items for clarity and consistency


Version.h
```diff
@@ -3,4 +3,4 @@
 // Also, numbers are interpreted as simple numbers.                        _   __   _
 // So 1.8 is actually 1.08, meaning that 1.12 is a later version than 1.8.  \_(..)_/
 
-#define VERSION "V1.11.9"
+#define VERSION "V1.13.1"
```
julianneswinoga: Could you explain the version bump? 13 vs 12, and why a minor and not just patch?

src/Mount.cpp
```diff
@@ -3306,13 +3322,20 @@ void Mount::moveSteppersTo(float targetRASteps, float targetDECSteps)
...
+        }
         targetDECSteps = min(targetDECSteps, (float) _decUpperLimit);
-        LOG(DEBUG_STEPPERS, "[STEPPERS]: MoveSteppersTo: DEC Upper Limit enforced. To: %f", targetDECSteps);
     }
+
     if (_decLowerLimit != 0)
     {
+        if (targetDECSteps < (float) _decLowerLimit)
+        {
+            LOG(DEBUG_STEPPERS, "[STEPPERS]: MoveSteppersTo: DEC Lower Limit enforced. To: %f", _decLowerLimit);
```
julianneswinoga: `%f` should be `%l`

src/c65_startup.hpp
```diff
@@ -57,28 +69,68 @@ bool processStartupKeys()
     bool waitForRelease = false;
     switch (startupState)
```
julianneswinoga, line 70: Just want to double check that you've tested the LCD with gyro level and auto-home. (I know you usually don't have an LCD)

julianneswinoga (dismissed):

Version.h
ClutchplateDude: I had 12 in another branch, but I guess I could go with 12 here. Given that it has some new features I thought a minor number change was appropriate

src/Mount.cpp
ClutchplateDude: Oooh, good catch!

src/c65_startup.hpp
ClutchplateDude: Actually, that's one combo I have not tested. Will do so today.

Version.h
julianneswinoga: Yeah I agree, large change but just wanted to make sure we were thinking the same thing. Minor version 12 makes sense. Keep it in order

Version.h
```diff
@@ -3,4 +3,4 @@
 // Also, numbers are interpreted as simple numbers.                        _   __   _
 // So 1.8 is actually 1.08, meaning that 1.12 is a later version than 1.8.  \_(..)_/
 
-#define VERSION "V1.11.9"
+#define VERSION "V1.12.1"
```
julianneswinoga: Looking at our previous version bumps, this should be 1.12.0 (not 12.1)

julianneswinoga (changes requested):

src/c65_startup.hpp
ClutchplateDude: I can't rewire my gyro level....

julianneswinoga: If we're making changes here but can't test them, I'm not sure we should be making those changes

julianneswinoga (changes requested):
Just need someone to test the LCD + gyro combination if we want those changes. Perhaps someone from the discord community would be willing?

src/c65_startup.hpp
ClutchplateDude: So basically stop development on LCD code?

src/Mount.cpp
```diff
@@ -1299,14 +1301,17 @@ void Mount::startSlewingToTarget()
...
     if ((_stepperRA->distanceToGo() != 0) || (_stepperDEC->distanceToGo() != 0))
     {
-        // Only stop tracking if we're actually going to slew somewhere else, otherwise the
+        // Only stop tracking if we're tracking and actually going to slew somewhere else, otherwise the
         // mount::loop() code won't detect the end of the slewing operation...
-        LOG(DEBUG_STEPPERS, "[MOUNT]: Stop tracking (NEMA steppers)");
-        stopSlewing(TRACKING);
-        _trackerStoppedAt        = millis();
-        _compensateForTrackerOff = true;
+        if (slewStatus() == SLEWING_TRACKING)
```
andre-stefanov, line 1367: `isSlewingTRK()` ?

andre-stefanov (changes requested):

src/Mount.cpp
ClutchplateDude: Yes

PR #197, opened 2022-10-10
Dew heater

anjok:
Adds dew heater support which can be adjusted from 0 10, stores values and restarts the heaters on reboot.

Adds `:XGh#` for getting and `:XShNnnn` for setting the values.

### DIFF
--- a/Configuration_adv.hpp
+++ b/Configuration_adv.hpp
@@ -537,7 +537,7 @@
 #if DEW_HEATER == 1
     // Careful with those values, they might melt your lens.
     // Make doubly sure your setup can handle 12V.
-    // this is the PWM duty cycle with vales from 0...255
+    // this is the PWM duty cycle with values from 0...255
     #if !defined(DEW_HEATER_1_MAX)
         #define DEW_HEATER_1_MAX 255
     #endif
--- a/Configuration_adv.hpp
+++ b/Configuration_adv.hpp
@@ -534,6 +534,18 @@
     #endif
 #endif
 
+#if DEW_HEATER == 1
+    // Careful with those values, they might melt your lens.
+    // Make doubly sure your setup can handle 12V.
+    // this is the PWM duty cycle with vales from 0...255
+    #if !defined(DEW_HEATER_1_MAX)
+        #define DEW_HEATER_1_MAX 255
+    #endif
+    #if !defined(DEW_HEATER_2_MAX)
+        #define DEW_HEATER_2_MAX 255
+    #endif
+#endif
+
 ////////////////////////////
 //
 // DEBUG OUTPUT
--- a/Configuration_adv.hpp
+++ b/Configuration_adv.hpp
@@ -535,7 +535,9 @@
 #endif
 
 #if DEW_HEATER == 1
-    // Careful with those values, they might melt your lens.
+    // Careful with those values, they might melt your lens
+    // or set your house on fire.
+    // Only use heaters with fuses!
     // Make doubly sure your setup can handle 12V.
     // this is the PWM duty cycle with values from 0...255
     #if !defined(DEW_HEATER_1_MAX)
--- a/boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
+++ b/boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
@@ -131,10 +131,10 @@
 #endif
 
 //Pin to turn on dew heater MOSFET
-#ifndef DEW_HEATER_PIN1
+#ifndef DEW_HEATER_1_PIN
     #define DEW_HEATER_1_PIN 10
 #endif
-#ifndef DEW_HEATER_PIN2
+#ifndef DEW_HEATER_2_PIN
     #define DEW_HEATER_2_PIN 7
 #endif
 
--- a/src/MeadeCommandProcessor.cpp
+++ b/src/MeadeCommandProcessor.cpp
@@ -820,15 +820,15 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
 //      Description:
 //        Get value for heaters
 //      Information:
-//        Get value for heater N in the range from 0-10, or both heaters if no N is given.
+//        Get value for heater N (0 or 1) as a value in the range from 0-10, or both heaters if no N is given.
 //      Returns:
-//        "NNN,NNN#" or "N#"
+//        "nn,nn#" or "n#"
 //
-// :XShNnnn#
+// :XShNnn#
 //      Description:
-//        Set strength for heater N to nnn
+//        Set strength for heater N to nn
 //      Information:
-//        Set scale of the PWM duty cycle for heater N (0 or 1) to nnn (0 - 10). The final PWM value goes from 0 to DEW_HEATER_N_MAX.
+//        Set scale of the PWM duty cycle for heater N (0 or 1) to nn (0 - 10). The final PWM value goes from 0 to DEW_HEATER_(N+1)_MAX.
 //      Returns:
 //        1#
 //
@@ -836,7 +836,7 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
 //      Description:
 //        Restart the mount
 //      Information:
-//        Restarts the mount like into a state which is like the one you get when you use the reset button.
+//        Restarts the mount in the same way pressing the Reset button on the LCD would.
 //      Returns:
 //        nothing, the mount reboots
 //
--- a/src/MeadeCommandProcessor.cpp
+++ b/src/MeadeCommandProcessor.cpp
@@ -1810,7 +1810,7 @@ String MeadeCommandProcessor::handleMeadeExtraCommands(String inCmd)
             {
                 // :XSh010# :XSh05# :XSh00#
                 // :XSh110# :XSh15# :XSh10#
-                // :XGh# :XGh0# :XGh1#
+                // :XGh# :XGh0# :XGh1# :XGh2#
                 _mount->setHeater(num, inCmd.substring(3).toInt());
                 return String("1#");
             }
--- a/src/MeadeCommandProcessor.cpp
+++ b/src/MeadeCommandProcessor.cpp
@@ -816,6 +816,22 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
 //      Returns:
 //        "HHMMSS"
 //
+// :XGh#
+//      Description:
+//        Get Heater values
+//      Information:
+//        Get the current values for both heaters
+//      Returns:
+//        "NNN,NNN"
+//
+// :XShNnnn#
+//      Description:
+//        Set strength for heater N to nnn
+//      Information:
+//        Set PWM duty cycle for heater N (0 or 1) to nnn (0 - 10). The final PWM value is goes from 0 to DEW_HEATER_N_MAX.
+//      Returns:
+//        1#
+//
 // :XSBn#
 //      Description:
 //        Set Backlash correction steps
@@ -1071,6 +1087,12 @@ String MeadeCommandProcessor::handleMeadeInit(String inCmd)
     _lcdMenu->printMenu("Remote control");
     _lcdMenu->setCursor(0, 1);
     _lcdMenu->printMenu(">SELECT to quit");
+
+#if DEW_HEATER == 1
+    LOG(DEBUG_ANY, "Starting dew heater");
+    _mount->heater(0, _mount->getHeater(0));
+    _mount->heater(1, _mount->getHeater(1));
+#endif
     return "";
 }
 
@@ -1669,13 +1691,21 @@ String MeadeCommandProcessor::handleMeadeExtraCommands(String inCmd)
 #if (WIFI_ENABLED == 1)
             return wifiControl.getStatus() + "#";
 #endif
-
             return "0,#";
+        }
+        else if (inCmd[1] == 'h')  // :XGh#
+        {
+#if (DEW_HEATER == 1)
+            char scratchBuffer[10];
+            sprintf(scratchBuffer, "%d,%d#", _mount->getHeater(0), _mount->getHeater(1));
+            return String(scratchBuffer);
+#endif
+            return "0,0#";
         }
     }
     else if (inCmd[0] == 'R')
     {
-        void(* resetFunc) (void) = 0;
+        void (*resetFunc)(void) = 0;
         resetFunc();
     }
 
@@ -1760,6 +1790,20 @@ String MeadeCommandProcessor::handleMeadeExtraCommands(String inCmd)
             {
                 _mount->setHomingOffset(StepperAxis::RA_STEPS, inCmd.substring(3).toInt());
             }
+        }
+        else if (inCmd[1] == 'h' && inCmd.length() > 3)
+        {
+#if (DEW_HEATER == 1)
+            long num = inCmd[2] - '0';
+            if (num >= 0 && num < 2)
+            {
+                // :XSh0255# :XSh05# :XSh128# :XSh00# :XSh1255# :XSh1128# :XSh10#
+                // :XSh0255# :XSh00# :XSh01#    :XGh# :XSh010# :XSh011#
+                _mount->heater(num, inCmd.substring(3).toInt());
+                return String("1#");
+            }
+#endif
+            return String("0#");
         }
     }
     else if (inCmd[0] == 'L')
--- a/src/MeadeCommandProcessor.cpp
+++ b/src/MeadeCommandProcessor.cpp
@@ -830,7 +830,7 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
 //      Description:
 //        Set strength for heater N to nn
 //      Information:
-//        Set scale of the PWM duty cycle for heater N to nn. 
+//        Set scale of the PWM duty cycle for heater N to nn.
 //        The final PWM value goes from 0 to DEW_HEATER_(N+1)_MAX.
 //      Parameters:
 //        "N" heater number is one of 0 or 1
--- a/src/MeadeCommandProcessor.cpp
+++ b/src/MeadeCommandProcessor.cpp
@@ -830,7 +830,8 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
 //      Description:
 //        Set strength for heater N to nn
 //      Information:
-//        Set scale of the PWM duty cycle for heater N to nn. The final PWM value goes from 0 to DEW_HEATER_(N+1)_MAX.
+//        Set scale of the PWM duty cycle for heater N to nn. 
+//        The final PWM value goes from 0 to DEW_HEATER_(N+1)_MAX.
 //      Parameters:
 //        "N" heater number is one of 0 or 1
 //        "nn" value in the range of 0 to 10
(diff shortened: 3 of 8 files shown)


src/MeadeCommandProcessor.cpp
```diff
@@ -816,6 +816,22 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
 //      Returns:
 //        "HHMMSS"
 //
+// :XGh#
+//      Description:
+//        Get Heater values
+//      Information:
+//        Get the current values for both heaters
+//      Returns:
+//        "NNN,NNN"
```
ClutchplateDude: Can we make the get and set functions symmetrical, i.e. have the get also use a N to signify which heater? You can make the number optional to save a call. So support both `:XGh#` and `:XGh1#`.

Configuration_adv.hpp
```diff
@@ -534,6 +534,18 @@
     #endif
 #endif
 
+#if DEW_HEATER == 1
```
ClutchplateDude, line 537: Let's use the usual `USE_XXXX` format, so rename this to `USE_DEW_HEATER`. Also, I think it might make sense to split heaters, or do we anticipate everyone having both heaters?

src/MeadeCommandProcessor.cpp
```diff
@@ -816,6 +816,22 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
...
+//
+// :XShNnnn#
+//      Description:
+//        Set strength for heater N to nnn
+//      Information:
+//        Set PWM duty cycle for heater N (0 or 1) to nnn (0 - 10). The final PWM value is goes from 0 to DEW_HEATER_N_MAX.
```
ClutchplateDude: Typo

src/MeadeCommandProcessor.cpp
```diff
@@ -1071,6 +1087,12 @@ String MeadeCommandProcessor::handleMeadeInit(String inCmd)
     _lcdMenu->printMenu("Remote control");
     _lcdMenu->setCursor(0, 1);
     _lcdMenu->printMenu(">SELECT to quit");
+
```
ClutchplateDude, line 1109: This doesn't seem like the right place for this code. It should be in `b_setup.hpp` around line 360

src/MeadeCommandProcessor.cpp
```diff
@@ -1669,10 +1691,24 @@ String MeadeCommandProcessor::handleMeadeExtraCommands(String inCmd)
...
+        {
+#if (DEW_HEATER == 1)
+            char scratchBuffer[10];
+            sprintf(scratchBuffer, "%d,%d#", _mount->getHeater(0), _mount->getHeater(1));
+            return String(scratchBuffer);
+#endif
+            return "0,0#";
+        }
     }
+    else if (inCmd[0] == 'R')
```
ClutchplateDude: Please add an entry to the docs area above, detailing this command.

src/Mount.cpp
```diff
@@ -3799,3 +3811,54 @@ void Mount::checkRALimit()
...
+    }
+    else if (num == 1)
+    {
+        pin = DEW_HEATER_2_PIN;
+        max = DEW_HEATER_2_MAX;
+    }
+    if (pin > 0)
+    {
+        val = map(val > 10 ? 10 : val, 0, 10, 0, max);
+        LOG(DEBUG_ANY, "[MOUNT]: heater %d (pin %d) setting to %d", num, pin, val);
```
ClutchplateDude: Let's scope the logs to `DEBUG_GENERAL`.

src/Mount.cpp
```diff
@@ -3799,3 +3811,54 @@ void Mount::checkRALimit()
...
     _lastTRKCheck = millis();
 }
+
+#if DEW_HEATER == 1
+/////////////////////////////////
+//
+// heater
+//
+/////////////////////////////////
+void Mount::heater(unsigned num, unsigned val)
```
ClutchplateDude: Can we name this function `setHeater()`, since the getter is called `getHeater()`.

ClutchplateDude (changes requested):
Thanks for your submission!

Configuration_adv.hpp
anjok: As there might be users who use DEW_HEATER already, I don't want to do that. Also, it's in a ton of places.

And no, it doesn't make sense. If you don't want two heaters, just don't set the value to the second one.

src/MeadeCommandProcessor.cpp
anjok: I actually did it that way because you had it like that for the XDGL...

src/MeadeCommandProcessor.cpp
anjok: The idea is that you don't want to heat until you're connected, in particular when you only start the mount.

CTetford: This is the opposite as to how the system has operated previously, it's common that people want it to start immediately since you could be setting up the OAT hours before astrodark to allow everything to acclimatize to ambient.

At the very least there needs to be a config option to start heating at boot using the previous EEPROM setting, or a configured default in the case of a cleared memory.

src/Mount.cpp
```diff
@@ -3799,3 +3811,54 @@ void Mount::checkRALimit()
...
+//
+// setHeater
+//
+/////////////////////////////////
+void Mount::setHeater(unsigned num, unsigned val)
+{
+    unsigned pin = 0, max = 0;
+    if (num == 0)
+    {
+        pin = DEW_HEATER_1_PIN;
```
CTetford: This should either check to make sure DEW_HEATER_N_PIN is defined, or ConfigurationValidation should be modified to make sure both pins are defined if DEW_HEATER is enabled.

src/MeadeCommandProcessor.cpp
anjok: Fixed "is goes"

src/MeadeCommandProcessor.cpp
anjok: Ah ok. I'll move it then. I normally have the thing sitting here turned on and wouldn't want to heat the guider, but I can just turn it off.

src/Mount.cpp
anjok: DEW_HEATER_N_PIN is always set due to the way it's set up the board files?

CTetford: Ideally yes, but I can easily see someone breaking it with a custom pins file or something. Just a precaution.

src/MeadeCommandProcessor.cpp
anjok: changed to XRST and doc'd

anjok:
There's no "Artifacts" from the latest clang-format fail?

And running locally also doesn't work:

```
astroberry@astroberry:~/Projects/OpenAstroTracker-Firmware $ bash -c 'shopt -s nullglob globstar;GLOBIGNORE=./src/libs/TimerInterrupt/*; for i in ./{.,src/**,unit_tests,boards/**}/*.{c,cpp,h,hpp}; do clang-format -i $i; done'
YAML:10:18: error: invalid boolean
AlignOperands:   Align
                 ^~~~~
Error reading /home/astroberry/Projects/OpenAstroTracker-Firmware/././.clang-format: Invalid argument
YAML:10:18: error: invalid boolean
```

CTetford:
You can see the artifacts from the summary page here:
https://github.com/OpenAstroTech/OpenAstroTracker-Firmware/actions/runs/3221360717

It seems Github changed the UI a bit, that message needs to be updated - we just noticed it yesterday.

anjok:
So pushing again didn't produce a clang-format.

anjok:
OK,  I actually did that before, but again couldn't find it.. you should change that text to say:

Click on Details, then click on "Summary" on left side. At the bottom of the page, there is an "Artifacts" section where you can download "clang-format-diff.patch"

src/MeadeCommandProcessor.cpp
ClutchplateDude: I know, but I have realized this is not optimal :-) But supporting both is straightforward.

anjok: So we should be good now.

Configuration_adv.hpp
ClutchplateDude: I think there's at best a handful, so we're going to change that regardless and just tell people, maybe we'll have a warning in the code that checks for the existence of that variable. And there is global find and replace so it doesn't really matter how many times it used, lol. Anyway, if you don't want to do it as part of this PR, that's fine, I can change it later for the next release.

src/MeadeCommandProcessor.cpp
ClutchplateDude: I don't see the optional character documented....

src/MeadeCommandProcessor.cpp
```diff
@@ -816,22 +816,30 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
...
 //        "HHMMSS"
 //
-// :XGh#
+// :XGhN#
 //      Description:
-//        Get Heater values
+//        Get value for heaters
 //      Information:
-//        Get the current values for both heaters
+//        Get value for heater N in the range from 0-10, or both heaters if no N is given.
```
ClutchplateDude: If you want to support more than 2 heaters, you'll need to adapt the EEPROM code accordingly. Also this needs to be limited to 9, since otherwise the setter can't figure out whether it's one or two digits (currently it checks for one only)

src/MeadeCommandProcessor.cpp
```diff
@@ -816,22 +816,30 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
...
-//        Set PWM duty cycle for heater N (0 or 1) to nnn (0 - 10). The final PWM value is goes from 0 to DEW_HEATER_N_MAX.
+//        Set scale of the PWM duty cycle for heater N (0 or 1) to nnn (0 - 10). The final PWM value goes from 0 to DEW_HEATER_N_MAX.
 //      Returns:
 //        1#
 //
+// :XRST#
+//      Description:
+//        Restart the mount
+//      Information:
+//        Restarts the mount like into a state which is like the one you get when you use the reset button.
```
ClutchplateDude: Too many 'like's. Let's do `Restarts the mount in the same way pressing the Reset button on the LCD would.`

src/MeadeCommandProcessor.cpp
```diff
@@ -1797,9 +1803,10 @@ String MeadeCommandProcessor::handleMeadeExtraCommands(String inCmd)
             long num = inCmd[2] - '0';
             if (num >= 0 && num < 2)
```
ClutchplateDude, line 1819: This should be `if (num >=0 && num<= 9)` now if you want to support more than 2. But then `setHeater()` code will currently crash for `num>1` since `_heaters` is defined with size 2. So it seems you are only supporting two, so the doc should reflect that.

ClutchplateDude (changes requested):

src/MeadeCommandProcessor.cpp
anjok: This is a misunderstanding. There's 2 heaters and the range they can operate in is 0 (off, 0V) to 10 (max 12V or 12V scaled to whatever you set the max to). So you'd get back `(4,8)` for example.

src/MeadeCommandProcessor.cpp
anjok: See above.

src/MeadeCommandProcessor.cpp
anjok: //        Get value for heater N (0 or 1) as a value in the range from 0-10, or both heaters if no N is given.

julianneswinoga:
> And running locally also doesn't work:
> 
> ```
> astroberry@astroberry:~/Projects/OpenAstroTracker-Firmware $ bash -c 'shopt -s nullglob globstar;GLOBIGNORE=./src/libs/TimerInterrupt/*; for i in ./{.,src/**,unit_tests,boards/**}/*.{c,cpp,h,hpp}; do clang-format -i $i; done'
> YAML:10:18: error: invalid boolean
> AlignOperands:   Align
>                  ^~~~~
> Error reading /home/astroberry/Projects/OpenAstroTracker-Firmware/././.clang-format: Invalid argument
> YAML:10:18: error: invalid boolean
> ```

What version of `clang-format` do you have installed? You need at least version 12 (from `README.md`)
```
$ clang-format --version
Ubuntu clang-format version 14.0.6-++20220622053019+f28c006a5895-1~exp1~20220622173056.159
```

julianneswinoga:
@anjok Please do not resolve issue conversations yourself, it goes against our [contribution guidelines](https://github.com/OpenAstroTech/OpenAstroTracker-Firmware#contribution):

> - The pull request can only be merged after all comments were resolved BY OAT DEVELOPERS. Please don't resolve the comments yourself since this can lead to missed issues.

Configuration_adv.hpp
```diff
@@ -534,6 +534,18 @@
...
+    // Careful with those values, they might melt your lens.
+    // Make doubly sure your setup can handle 12V.
+    // this is the PWM duty cycle with vales from 0...255
+    #if !defined(DEW_HEATER_1_MAX)
+        #define DEW_HEATER_1_MAX 255
```
julianneswinoga, line 544: These values should not default to full power. Maybe something like 10%?

Configuration_adv.hpp
```diff
@@ -590,4 +602,4 @@
     #else
         #error "Debugging not supported on this platform"
     #endif
-#endif
+#endif
```
julianneswinoga: Please add back the newline at the end of the file

boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
```diff
@@ -131,10 +131,10 @@
 #endif
 
 //Pin to turn on dew heater MOSFET
-#ifndef DEW_HEATER_PIN1
+#ifndef DEW_HEATER_1_PIN
```
julianneswinoga: If we're fixing these we might as well fix the rest of the pins files, not just for MKS v1

src/Mount.cpp
```diff
@@ -3023,13 +3033,15 @@ void Mount::setDecLimitPositionAbs(bool upper, long stepperPos)
 {
     if (upper)
     {
-        _decUpperLimit = DEC_LIMIT_UP * _stepsPerDECDegree;
+        stepperPos     = stepperPos == 0 ? DEC_LIMIT_UP * _stepsPerDECDegree : stepperPos;
```
julianneswinoga: Why are we changing DEC limits in a dew heater PR...?

src/Mount.cpp
```diff
@@ -3799,3 +3811,64 @@ void Mount::checkRALimit()
...
+    }
+    else
+    {
+        LOG(DEBUG_MOUNT, "[MOUNT]: wrong pin number %d", num);
+    }
+}
+
+unsigned Mount::getHeater(unsigned num)
+{
+    unsigned val = _heaters[num], max;
```
julianneswinoga: Please split up this declaration, it's a bit confusing to read like this

anjok:
I simply installed with `sudo apt-get install clang-format`, there apparently is no version 12.

```
astroberry@astroberry:~ $ clang-format --version
clang-format version 7.0.1-8+rpi3+deb10u2 (tags/RELEASE_701/final)
```

And for the Mac:

```
anjo:OpenAstroTracker-Firmware akrank$ bash -c 'shopt -s nullglob globstar;GLOBIGNORE=./src/libs/TimerInterrupt/*; for i in ./{.,src/**,unit_tests,boards/**}/*.{c,cpp,h,hpp}; do clang-format -i $i; done'
bash: line 0: shopt: globstar: invalid shell option name
```

So neither work.

julianneswinoga (changes requested):

boards/AVR_MKS_GEN_L_V1/pins_MKS_GEN_L_V1.h
julianneswinoga: Ah nevermind just saw these are all fixed in #196 . So lets get that in and then this work will be on top of that.

Configuration_adv.hpp
CTetford: I disagree, I think this approach is correct - the max value should be 255, the default operating setting is what should be lower.

anjok: It's what is currently used and if people have 12v heaters it should be fine. if you have 5v you need to test anyway.

src/Mount.cpp
anjok: because that's what i work in and it's still not fixed *weeks* after the break in 1.11.9. If you want to gone, just apply the commit into develop.

anjok:
https://apple.stackexchange.com/questions/291287/globstar-invalid-shell-option-name-on-macos-even-with-bash-4-x

This seems to work for Mac OS. You might want to put this in the description.

src/MeadeCommandProcessor.cpp
```diff
@@ -1693,17 +1696,20 @@ String MeadeCommandProcessor::handleMeadeExtraCommands(String inCmd)
 #endif
             return "0,#";
         }
-        else if (inCmd[1] == 'h')  // :XGh#
+        else if (inCmd[1] == 'h')  // :XGh# :XGh0#
         {
 #if (DEW_HEATER == 1)
             char scratchBuffer[10];
-            sprintf(scratchBuffer, "%d,%d#", _mount->getHeater(0), _mount->getHeater(1));
+            if (inCmd.length() > 2)
```
ClutchplateDude, line 1714: Curly braces are required around every conditional body.

src/MeadeCommandProcessor.cpp
ClutchplateDude: Oh, ok. The wording is ambiguous.
`Get value for heater N (either 0 or 1) or both if no N is given. Returned value is in the range 0-10.`

src/MeadeCommandProcessor.cpp
```diff
@@ -816,22 +816,30 @@ bool gpsAqcuisitionComplete(int &indicator);  // defined in c72_menuHA_GPS.hpp
...
+// :XGhN#
 //      Description:
-//        Get Heater values
+//        Get value for heaters
 //      Information:
-//        Get the current values for both heaters
+//        Get value for heater N in the range from 0-10, or both heaters if no N is given.
 //      Returns:
-//        "NNN,NNN"
+//        "NNN,NNN#" or "N#"
```
ClutchplateDude: `"NN,NN# or "NN# where NN is the power the heaters are set to in the range 0 (off) to 10 (full power).`

ClutchplateDude (changes requested):

src/Mount.cpp
ClutchplateDude: A fix for this has just been merged into `develop`. You can remove it from this PR.

ClutchplateDude:
I'm still waiting on you to make the requested changes....

src/Mount.cpp
```diff
@@ -2699,6 +2708,16 @@ void Mount::interruptLoop()
         _stepperFocus->runSpeed();
     }
 #endif
+#if DEW_HEATER == 1
+    if(_heaters[0]) 
+    {
+        _heaters[0]->runLoop();
```
CTetford, line 2714: From what you mentioned on Discord, the analogWrite PWM output works on pin D7 but not D10. Rather than implementing a pseudo-PWM system within the loop, why not have one port using analogWrite for PWM (D7) dew heater control and the other just on/off (D10)? If you have two heaters that require PWM output, they can be connected to the same port.

We'd rather not add extra complication/overhead to the main loop to support a relatively minor feature.

PR #200, opened 2022-10-16
Dec limit fix to return degrees

ClutchplateDude:
Return degrees in limit query and support returning each one individually.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,6 @@
+**V1.11.12 - Updates**
+- Fix DEC limit code to return degrees instead of stepper positions. Support separate limits being queried.
+
 **V1.11.11 - Updates**
 - Fix DEC limit code to use parameters.
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,6 @@
+**V1.11.11 - Updates**
+- Fix DEC limit code to use parameters.
+
 **V1.11.10 - Updates**
 - Fix dew heater pin assignments.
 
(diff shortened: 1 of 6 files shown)


src/c78_menuINFO.hpp
```diff
@@ -136,17 +136,17 @@ void printStatusSubmenu()
...
                 }
                 break;
 
             case 6:
                 {
-                    long lowerLimit, upperLimit;
+                    float lowerLimit, upperLimit;
                     mount.getDecLimitPositions(lowerLimit, upperLimit);
-                    lcdMenu.printMenu("DEC Up: " + String(upperLimit));
+                    lcdMenu.printMenu("DEC Up: " + String(upperLimit, 1));
```
julianneswinoga, line 149: Base 1? Am I missing something here?

julianneswinoga:
Lets also make sure to regenerate the documentation here after it's merged: https://wiki.openastrotech.com/en/Knowledge/Firmware/MeadeCommands

Changelog.md
```diff
@@ -1,3 +1,6 @@
+**V1.11.12 - Updates**
+- Fix DEC limit code to return degrees instead of stepper positions. Support separate limits being queried.
```
julianneswinoga: IMO also reword to say "Change" instead of "Fix", since it's an interface change not a bug

julianneswinoga (changes requested):

src/c78_menuINFO.hpp
julianneswinoga: Ah read the wrong implementation. For floats it's the number of decimal places :woman_facepalming:

julianneswinoga (approved):

PR #201, opened 2022-11-04
End switches for RA and/or DEC

ClutchplateDude:
Switches to stop slewing and tracking when running into end switches. Currently only supported on MKS21.


src/EndSwitches.cpp
```diff
@@ -0,0 +1,202 @@
...
+    _minPin           = minPin;
+    _maxPin           = maxPin;
+    _posWhenTriggered = 0;
+    pinMode(_minPin, INPUT_PULLUP);
+    pinMode(_maxPin, INPUT_PULLUP);
+}
+
+/////////////////////////////////
+//
+// getSwitchState
```
julianneswinoga: Can we remove the docblocks that don't have any actual documentation in them? It's just fluff

src/EndSwitches.cpp
```diff
@@ -0,0 +1,202 @@
...
+//
+// getPosWhenTriggered
+//
+/////////////////////////////////
+long EndSwitch::getPosWhenTriggered() const
+{
+    return _posWhenTriggered;
+}
+
+/////////////////////////////////
```
julianneswinoga: We use currently use `/**` style comments more, lets stick to that style

src/EndSwitches.cpp
```diff
@@ -0,0 +1,202 @@
...
+// so it needs to be very fast. We only set a flag and remember
+// when we triggered the switch.
+//
+/////////////////////////////////
+void EndSwitch::processEndSwitchState()
+{
+    switch (_state)
+    {
+        case EndSwitchState::SWITCH_NOT_ACTIVE:
+            {
```
julianneswinoga, line 43: You don't need the braces inside the `case` statement (only needed if you're defining variables)

src/EndSwitches.cpp
```diff
@@ -0,0 +1,202 @@
...
+///////////////////////////
+//
+// checkSwitchState
+// This runs in the Mount::loop() function where all the logic
+// about stopping, parking, etc. happens.
+//
+///////////////////////////
+void EndSwitch::checkSwitchState()
+{
+    if ((_state == EndSwitchState::SWITCH_AT_MINIMUM) || (_state == EndSwitchState::SWITCH_AT_MAXIMUM))
```
julianneswinoga: You can flip the conditional around and make it early return. Reduces the nesting logic
```C++
if (!(_state == EndSwitchState::SWITCH_AT_MINIMUM || _state == EndSwitchState::SWITCH_AT_MAXIMUM))
{
    // Don't run any logic if we're not at the stops
    return;
}
```

src/EndSwitches.cpp
```diff
@@ -0,0 +1,202 @@
...
+                "[ENDSWITCH]: Reached maximum position on %s axis while tracking. Stopping tracking. State is now SWITCH_SLEWING_OFF",
+                _axis == StepperAxis::RA_STEPS ? "RA" : "DEC");
+            _pMount->stopSlewing(TRACKING);
+            _state = EndSwitchState::SWITCH_SLEWING_OFF_MAXIMUM;
+        }
+        else if (_state == EndSwitchState::SWITCH_AT_MINIMUM)
+        {
+            // This shouldn't really happen, since we can't really reach minimum if we're not slewing, since tracking goes to maximum.
+            // But just so that the state machine resets correctly once we move off this switch we transition to the state that waits for that.
+            _state = EndSwitchState::SWITCH_SLEWING_OFF_MINIMUM;
```
julianneswinoga: If this is an invalid state, adding a log would be helpful here

src/EndSwitches.hpp
```diff
@@ -0,0 +1,51 @@
...
+    int _dir;
+    int _minPin;
+    int _maxPin;
+
+  public:
+    EndSwitch(Mount *mount, StepperAxis axis, int minPin, int maxPin, int activeState);
+    void processEndSwitchState();
+    void checkSwitchState();
+
+    EndSwitchState getSwitchState() const;
```
julianneswinoga: These 3 methods aren't used, I would remove them

src/Mount.cpp
```diff
@@ -2635,7 +2661,8 @@ void Mount::delay(int ms)
 /////////////////////////////////
 void Mount::interruptLoop()
 {
-    if (_mountStatus & STATUS_GUIDE_PULSE)
+    // Only process guide pulses if we are tracking.
+    if ((_mountStatus & STATUS_GUIDE_PULSE) && (_mountStatus & STATUS_TRACKING))
```
julianneswinoga, line 2677: I'd add this change to the changelog

src/Types.hpp
```diff
@@ -0,0 +1,13 @@
+#ifndef _TYPES_HPP_
```
julianneswinoga, line 1: If we're moving generic types here, lets also move `LocalDate` and `PointOfInterest`

src/Mount.hpp
```diff
@@ -343,6 +336,11 @@ class Mount
     // Set the speed of the given motor
     void setSpeed(StepperAxis which, float speedDegsPerSec);
 
+#if (USE_RA_END_SWITCH == 1) || (USE_DEC_END_SWITCH == 1)
+    void setupEndSwitches();
+    void processEndSwitchState();
```
julianneswinoga: `processEndSwitchState` isn't used (and not implemented)

julianneswinoga (changes requested):

src/EndSwitches.cpp
ClutchplateDude: I'd rather not. They make it real easy to scan functions

src/EndSwitches.cpp
ClutchplateDude: I don't see that style in the few files I looked at... they all had the ///// style.

src/EndSwitches.cpp
ClutchplateDude: I know, but the consistency of either having them everywhere or nowhere makes it much easier to parse and see the structure

src/EndSwitches.cpp
ClutchplateDude: I'm not a huge fan of early returns, but ok.

src/EndSwitches.cpp
ClutchplateDude: Yeah.

src/EndSwitches.hpp
ClutchplateDude: Good catch.

src/Mount.cpp
ClutchplateDude: yup

src/Mount.hpp
ClutchplateDude: Removed.

src/Types.hpp
ClutchplateDude: Separate PR for that.

src/EndSwitches.cpp
julianneswinoga: Huh, checked again and you're right. Not sure how I was searching before.

src/EndSwitches.cpp
julianneswinoga: In most cases early returns reduce the number of states that you need to keep in your head, makes reading code easier: https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement
(same concept as const-qualifying everything)

src/Types.hpp
julianneswinoga: Fair

src/EndSwitches.cpp
julianneswinoga: IMO that sounds like an editor preference and not something that should be in the code.
Here's a plugin for VSCode: https://marketplace.visualstudio.com/items?itemName=alefragnani.separators (i.e. platformio IDE) and CLion/Intellij has it built in: https://www.jetbrains.com/webstorm/guide/tips/method-separators/

ClutchplateDude: Pretty much the entire codebase is like this. I'm not making this file different. And making it editor dependent defeats the purpose of making code browsable, regardless of where.

julianneswinoga (approved):

src/EPROMStore.cpp
```diff
@@ -414,7 +414,7 @@ void EEPROMStore::storeRAStepsPerDegree(float raStepsPerDegree)
     // Store steps as 100x steps/deg at 256 MS.
     const float factor = SteppingStorageNormalized / RA_TRACKING_MICROSTEPPING;
     int32_t val        = raStepsPerDegree * factor;
-    LOG(DEBUG_EEPROM, "[EEPROM]: Storing RA steps to %d (%f)", val, raStepsPerDegree);
+    LOG(DEBUG_EEPROM, "[EEPROM]: Storing RA steps to %l (%f)", val, raStepsPerDegree);
```
andre-stefanov, line 417: i think this should be `%ld`

PR #204, opened 2022-11-28
V1.11.14 - Updates

ClutchplateDude:
- Prevented firmware from hanging at boot if LCD is defined but not connected.
- Fixed a (rare) bug that would throw off GoTo commands when a target coordinate had been set.

### DIFF
--- a/src/Mount.cpp
+++ b/src/Mount.cpp
@@ -1035,9 +1035,10 @@ String Mount::getMountHardwareInfo()
     #if (USE_RA_END_SWITCH == 1)
     ret += F("_RA");
     #endif
-    #if || (USE_DEC_END_SWITCH == 1)
+    #if (USE_DEC_END_SWITCH == 1)
     ret += F("_DEC");
     #endif
+    ret += F(",");
 #else
     ret += F("NO_ENDSW,");
 #endif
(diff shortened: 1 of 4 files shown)


src/b_setup.hpp
```diff
@@ -232,18 +232,31 @@ void setup()
...
-        LOG(DEBUG_INFO, "[SYSTEM]: Waiting for button release!");
+        LOG(DEBUG_INFO, "[SYSTEM]: DOWN pressed, waiting for button release!");
+        bool timedOut = false;
         while (lcdButtons.currentState() != btnNONE)
         {
             delay(10);
+            // Wait 3s maximum for button to be released.
+            if (millis() - lcdCheckStart > 3000)
+            {
+                LOG(DEBUG_INFO, "[SYSTEM]: Timedout waiting for button release. LCD not installed?");
```
julianneswinoga, line 249: Could we do something like force `lcdButtons.currentState()` to always return `btnNONE` if we get into this state? Just to enforce that nothing else in the FW could be waiting on the lcd

julianneswinoga (changes requested):
Great stuff! Was this tested with/without LCD? :smile:

src/b_setup.hpp
ClutchplateDude: I did not test without LCD.... Maybe someone else can.

ClutchplateDude: This was tested with and without LCD enabled on my second controller that does not have an LCD attached. Works as designed.

src/Mount.cpp
```diff
@@ -3429,7 +3430,7 @@ void Mount::moveStepperBy(StepperAxis direction, long steps)
             }
             break;
         case DEC_STEPS:
-            moveSteppersTo(_stepperRA->targetPosition(), _stepperDEC->targetPosition() + steps);
+            moveSteppersTo(_stepperRA->currentPosition(), _stepperDEC->currentPosition() + steps);
```
andre-stefanov, line 3433: isn't this an ra->stop and dec->move? would be easier to read/understand.

andre-stefanov (changes requested):

src/Mount.cpp
ClutchplateDude: moveSteppersBy() is only called by the MX command. and the homing and endswitch code, which both stop the motors before moving them. Using target would be incorrect for them, so current is correct.

andre-stefanov (approved):

PR #205, opened 2022-11-30
added config version validation

### LINKED ISSUE #203: Introduce configuration versioning
we should add a version value to each generated local config, e.g:

```cpp
#define LOCAL_CONFIG_VERSION 1
```

the firmware code could then have following definition & check:

```cpp
#define MIN_LOCAL_CONFIG_VERSION 1
#if MIN_LOCAL_CONFIG_VERSION > LOCAL_CONFIG_VERSION
  #error You have to update your local config to be able to use this version of firmware
#endif
```

This would allow us making changes to the config api (e.g. renaming or restructuring the config definitions) which would not lead to compile errors otherwise.
The user would also have an easier to understand error message instead of some generic "XYZ not defined"
If we introduce some breaking changes, we simply bump the min required config version in the firmware ... then adapt the config server and bump the version there as well.

andre-stefanov:


ConfigurationValidation.hpp
```diff
@@ -6,6 +6,12 @@
 //                            ////////
 //////////////////////////////////////
 
+#define MIN_CONFIG_VERSION 1
+
+#if CONFIG_VERSION < MIN_CONFIG_VERSION
```
ClutchplateDude: If CONFIG_VERSION is not defined will this still be true?

andre-stefanov: good catch, seems i forgot to commit this

PR #206, opened 2022-12-01
AVR interrupt based stepper

andre-stefanov:


src/DayTime.cpp
```diff
@@ -153,14 +153,14 @@ void DayTime::addHours(float deltaHours)
 
 void DayTime::checkHours()
 {
-    while (totalSeconds >= secondsPerDay)
+    while (totalSeconds >= SIDEREAL_SECONDS_PER_DAY)
```
ClutchplateDude: This must be reverted. This needs to be actual seconds, not sidereal.

ClutchplateDude (changes requested):

src/DayTime.cpp
andre-stefanov: done

ClutchplateDude (dismissed):
Great work! This is going to be amazing!

ClutchplateDude (approved):
LGTM

PR #207, opened 2022-12-05
V1.12.1 - Build break fix

ClutchplateDude:
- Fixed a bug that caused Autohoming to cause a build break.
- Fixed a bug that would prevent Hall sensor based Autohoming from completing.


julianneswinoga (changes requested):
Can you update `matrix_build.py` so that it catches this path in the future?

ClutchplateDude:
I'm not familiar with how all that stuff works.... Andre does most of that, I believe.... 
just trying to unblock people from failed builds right now.

ClutchplateDude:
Well that's going to balloon our build times :-)

PR #208, opened 2022-12-09
fixed steppers not stopping from setSpeed(0)

andre-stefanov:


julianneswinoga (approved):

src/InterruptAccelStepper.h
```diff
@@ -64,7 +64,15 @@ template <typename STEPPER> class InterruptAccelStepper
     {
         LOG(DEBUG_STEPPERS, "[IAS-%d] setSpeed(%f)", STEPPER::TIMER_ID, speed);
         _speed = speed;
-        STEPPER::moveTo(_speed, INT32_MAX);
+
+        if (speed == 0.0f)
```
julianneswinoga: Shouldn't compare`==`  with floats

julianneswinoga (changes requested):

src/InterruptAccelStepper.h
andre-stefanov: yeah i know, but to be honest i hoped that 0.0f should always be correct (at least if ignoring -0.0f) because sign, exponent, mantissa are zero for this special value. and thus should be comparable without issues. what do you think?

andre-stefanov: In theory we could use epsilon comparison here ... but not sure which epsilon value would be ok here.

julianneswinoga: Is there a slowest value that your library can drive steppers at? Maybe there's a timer limitation?
If not I'd just make up a slowest value and call that good (like 1 step a day or something crazy)

andre-stefanov:
Good idea. Will do that. Thank you!

Julianne Swinoga ***@***.***> schrieb am Fr., 9. Dez. 2022,
15:03:

src/InterruptAccelStepper.h
andre-stefanov: Good idea, will do that. Thank you!

andre-stefanov: done

ClutchplateDude (dismissed):

src/InterruptAccelStepper.h
```diff
@@ -64,7 +67,15 @@ template <typename STEPPER> class InterruptAccelStepper
     {
         LOG(DEBUG_STEPPERS, "[IAS-%d] setSpeed(%f)", STEPPER::TIMER_ID, speed);
         _speed = speed;
-        STEPPER::moveTo(_speed, INT32_MAX);
+
+        if (abs(speed) < MIN_STEPS_PER_SEC)
```
julianneswinoga: Why do you use Arduino's `abs()` but also you define `ABS()` in this file? Just kind of generally confused what's going on with the usage

andre-stefanov: to be honest atm i forgot why i defined my custom one. i think back then the arduino one could not be found for some stupid reason ... and this time i just used the arduino abs as usual

andre-stefanov: i removed the custom one

ClutchplateDude: Shouldn't it be fabs()?

andre-stefanov: well ... in this case then even `fabsf` (for float instead of double)

andre-stefanov: done

julianneswinoga (approved):

PR #211, opened 2022-12-13
Feature/js/parallelize matrix build

julianneswinoga:
Linking is a single-threaded task it does not effectively utilize multicore processors. By changing `matrix_build.py` to instead create N projects which each will build a solution in parallel we can avoid the single-core linking bottleneck. This scales very well with the number of processors. Building `mksgenlv21`:
|          Host          | old time | new time |
|------------------------|----------|----------|
| Personal server (N=40) | 30m23s   | 6m29s    |
| Laptop (N=8)           | 14m10s   | 6m54s    |

Our current GitHub Actions workflow only runs on dual core machines (N=2), so the speedup is marginal (but tangible) but this prepares us well for if we want to run the build process on a more powerful CI.
|    env     | old time | new time |
|------------|----------|----------|
| mksgenlv21 | 14m47s   | 9m55s    |
| mksgenlv2  | 18m10s   | 10m9s    |
| mksgenlv1  | 4m30s    | 3m13s    |
| esp32      | 3m32s    | 3m3s     |
| ramps      | 15m26s   | 9m55s    |


andre-stefanov (dismissed):

christian-kardach (dismissed):
Looks good to me!

PR #217, opened 2023-01-10
V1.12.6 - Southern Hemisphere fixes

ClutchplateDude:
- Fixed all known issues related to running in the southern hemisphere.
- Added support for DIR inversion to interrupt stepper library.
- Hemisphere no longer needs to be defined in config, firmware determines it automatically from the given Latitude, switching at the equator.
- Fixed the logic in Sync call to account for both hemispheres.
- Fixed some logic that stopped tracking when setting home position.
- Fixed a bug that caused the firmware to fail to recognize the end of very short slews.
- Added Meade Extension command to query remaining tracking time
- Added Meade Extension command to query the hemisphere that is set.


src/Declination.cpp
```diff
@@ -96,32 +99,33 @@ const char *Declination::ToString() const
 Declination Declination::ParseFromMeade(String const &s)
 {
     Declination result;
-    LOG(DEBUG_GENERAL, "[DECLINATION]: Declination.Parse(%s)", s.c_str());
+    LOG(DEBUG_GENERAL, "[DECLINATION]: Declination.Parse(%s) for %s Hemi", s.c_str(), inNorthernHemisphere ? "N" : "S");
```
andre-stefanov: since this is used for parsing from meade ... we probably should rather log it into `DEBUG_MEADE`

src/Mount.cpp
```diff
@@ -86,6 +86,10 @@ Mount::Mount(LcdMenu *lcdMenu)
 
 void Mount::initializeVariables()
 {
+    // We are now defaulting to northern hemisphere at 45deg. Switching is now supported
+    // at runtime when the Latitude is received via Meade command.
+    inNorthernHemisphere = true;  // Since we set latitude to +45 below.
```
andre-stefanov: we should allow changing the default (for standalone builds in SH)

src/Mount.cpp
```diff
@@ -206,6 +210,41 @@ void Mount::readPersistentData()
...
+{
+    if ((inNorthernHemisphere != inNorthern) || force)
+    {
+        bool wasTracking = isSlewingTRK();
+        LOG(DEBUG_ANY, "[SYSTEM]: Hemisphere changed (or forced update) to %s.", inNorthern ? "northern" : "southern");
+        LOG(DEBUG_ANY, "[SYSTEM]: Stopping all steppers.");
+        stopSlewing(ALL_DIRECTIONS | TRACKING);
+        waitUntilStopped(ALL_DIRECTIONS);
+        inNorthernHemisphere = inNorthern;
+        bool invertDir       = inNorthernHemisphere ? (RA_INVERT_DIR == 1) : (RA_INVERT_DIR != 1);
```
andre-stefanov, line 228: `bool invertDir = inNorthernHemisphere == RA_INVERT_DIR`

src/Mount.cpp
```diff
@@ -1317,16 +1375,16 @@ void Mount::startSlewingToTarget()
...
-
 #if DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART
     // Since normal state for DEC is guide microstepping, switch to slew microstepping here.
     LOG(DEBUG_STEPPERS, "[STEPPERS]: startSlewingToTarget: Switching DEC driver to microsteps(%d)", DEC_SLEW_MICROSTEPPING);
     _driverDEC->microsteps(DEC_SLEW_MICROSTEPPING == 1 ? 0 : DEC_SLEW_MICROSTEPPING);
 #endif
+    _stepperWasRunning = true;
+    moveSteppersTo(targetRAPosition, targetDECPosition, RA_AND_DEC_STEPS);  // u-steps (in slew mode)
+    _totalDECMove = 1.0f * _stepperDEC->distanceToGo();
+    _totalRAMove  = 1.0f * _stepperRA->distanceToGo();
```
andre-stefanov: static cast?

andre-stefanov (changes requested):

src/Declination.cpp
ClutchplateDude: Good point.

src/Mount.cpp
ClutchplateDude: I did this because I wanted to be sure that it will work since inNorthernHermisphere is a bool, but RA_INVERT_DIR is '0' or '1'... I don't think that 'true == 1' is true.

src/Mount.cpp
ClutchplateDude: Yup

src/Mount.cpp

src/Declination.cpp
ClutchplateDude: Changed a few more.

src/Mount.cpp
andre-stefanov: well in worst case we could static cast ... but i think it is fine this way

andre-stefanov (approved):

PR #228, opened 2023-05-23
V1.12.13 - Updates

ClutchplateDude:
- Added ability to override RA Wheel circumference in local configuration
- Added a physical limit define which defines how far RA can turn before physical issues arise.
- Manual slewing now targets the physical limit of the axis, instead of just going x thousand steps.
- Added support to define the back slew distance that is required to un-signal a signaled end switch. Some mechanical switches have a hysteresis that requires a lot (600%!) of back movement to un-signal.
- Manual slewing termination was incorrect and not executing pending operations.
- Set Tracking compensation speed to slewing speed instead of other calculation that was way too slow and caused long post-slew delays.
- Firmware returns"OpenAstroMount" for :GVP# Meade command if compiled for OAM.
- Enhanced logging output to show time since last log output.
- Added output of all stepper settings to log output.


Configuration_adv.hpp
```diff
@@ -474,25 +480,40 @@
 
 // RA EndSwitch support
 //////////////////////////////////////////
+// End switches allow you to define how many percent of the distance needed to stop
```
julianneswinoga: I would move this down to where `RA_ENDSWITCH_BACKSLEW_PERCENT` is defined (i.e. not at the top `USE_RA_END_SWITCH` block)

julianneswinoga: I'd also add something about what "percent" means. Like how does it correspond to physical movement? When first reading I expected it to be either steps or hours etc. Some sort of comment explanation would help a lot.

src/Utility.cpp
```diff
@@ -357,24 +357,30 @@ String format(const char *input, ...)
...
     #else
         DEBUG_SERIAL_PORT.print("[");
         DEBUG_SERIAL_PORT.print(String(now));
-        DEBUG_SERIAL_PORT.print("]:");
+        DEBUG_SERIAL_PORT.print("] ms:{");
+        DEBUG_SERIAL_PORT.print(String(delta));
+        DEBUG_SERIAL_PORT.print("} ms:");
         DEBUG_SERIAL_PORT.print(String(freeMemory()));
-        DEBUG_SERIAL_PORT.print(": ");
+        DEBUG_SERIAL_PORT.print(" B: ");
```
julianneswinoga: Why is there a `B` here?

src/StepperConfiguration.hpp
```diff
@@ -40,8 +40,8 @@ struct Ra {
     using stepper_slew = Stepper<interrupt, driver, ramp_slew>;
     using stepper_trk  = Stepper<interrupt, driver, ramp_trk>;
 
-    constexpr static float SPEED_COMPENSATION
-        = static_cast<float>(interrupt::FREQ) / static_cast<float>(ramp_slew::interval(1)) - (SPEED_TRK * 2.0f);
+    constexpr static float SPEED_COMPENSATION = SPEED_SLEW;
+    // = static_cast<float>(interrupt::FREQ) / static_cast<float>(ramp_slew::interval(1)) - (SPEED_TRK * 2.0f);
```
julianneswinoga: I don't understand why you're removing this?

src/Mount.cpp
```diff
@@ -2292,14 +2292,26 @@ void Mount::startSlewing(int direction)
 
             if (direction & EAST)
             {
-                LOG(DEBUG_STEPPERS, "[STEPPERS]: startSlewing(E): initial targetMoveTo is %l", -sign * 300000);
-                _stepperRA->moveTo(-sign * 300000);
+                // We need to subtract the distance tracked from the physical RA home coordinate
+                float trackedHours = (_stepperTRK->currentPosition() / _trackingSpeed) / 3600.0F;  // steps / steps/s / 3600 = hours
```
julianneswinoga: Pull `trackedHours` out of the `if`, make it `const`

julianneswinoga (changes requested):

src/Utility.cpp
ClutchplateDude: Indicates 'Bytes'

src/StepperConfiguration.hpp
ClutchplateDude: Because it doesn't work. It causes long delays after slews because it is way too low. And Andre has no time to investigate why or explain why this calculation is done the way it is.

Configuration_adv.hpp
ClutchplateDude: I mean that whole comment is a detailed explanation. Not sure what else I can add

src/Mount.cpp
ClutchplateDude: Good catch

src/Utility.cpp
julianneswinoga: Gotcha, I'd delete the space before then. `1024B` makes a bit more sense than `1024 B`

src/Utility.cpp
```diff
@@ -357,24 +357,30 @@ String format(const char *input, ...)
...
+        unsigned long delta = now - lastLog;
         va_list argp;
         va_start(argp, input);
     #if BUFFER_LOGS == true
         addToLogBuffer(formatArg(input.c_str(), argp));
     #else
         DEBUG_SERIAL_PORT.print("[");
         DEBUG_SERIAL_PORT.print(String(now));
-        DEBUG_SERIAL_PORT.print("]:");
+        DEBUG_SERIAL_PORT.print("] ms:{");
```
julianneswinoga: delete the pre `ms:` and the space
Something like this is a bit more compact and readable:
`[10058]{3}ms 1024B: [MOUNT] Test message`

src/StepperConfiguration.hpp
julianneswinoga: Gotcha, remove the commented out code then

Configuration_adv.hpp
julianneswinoga: Like how is the percentage calculated? Percent is a relative measure between two values - what values are those in this case? It's not clear to me

julianneswinoga (changes requested):

Configuration_adv.hpp
ClutchplateDude: Hmmm, ok let me see if I can explain it and then you can maybe word it better. 
When the switch triggers, it takes a certain distance until we stop (deceleration ramp). The number of steps needed to stop is the 100% value. We then slew in the opposite direction by this number of steps times the given percentage.

ClutchplateDude: We could also specify this in degrees from the trigger point. And that might actually make more sense, since then we are not dependent on the speed and thus the length of the ramp. Since the hysteresis of a switch is constant, degrees are probably better anyway. I'll change that.

julianneswinoga (approved):

PR #235, opened 2023-08-15
LCD menu updates (RA-DEC homing) and stop parking on setting home position via the LCD

Trigger-broom-289:
LCD menu updates (RA-DEC homing) and stop parking on setting home position via the LCD


Trigger-broom-289:
I meant to write: **stop tracking** instead of _stop parking_...

Trigger-broom-289:
> parking

tracking

PR #241, opened 2024-02-21
fix sign of 0 is ignored

peteasa:
I noticed that setting Latitude of observatories close to the meridian fails. The failure occurs because -0 degrees is not necessarily a negative value.

This draft change is for your consideration and fixes this issue.


src/Longitude.cpp
```diff
@@ -83,12 +83,17 @@ const char *Longitude::formatString(char *targetBuffer, const char *format, long
     secs      = secs - degs * 3600;
     long mins = secs / 60;
     secs      = secs - mins * 60;
+    char sgn;
     if (totalSeconds < 0)
     {
-        degs = -degs;
```
ClutchplateDude: I'm not sure this will work correctly. `formatStringImpl` makes decisions based on the sign of `degs`, which would now behave differently.... we may need to look at all calls to `formatString` and see whether they pass `+`, which invokes that test.

peteasa: formatStringImpl only makes the decision based on sign of degs if the char sgn = \0.. in my case I put a character so formatStringImpl simply copies sgn to the Meade command.

` if (sgn != '\0')
    {
        achDegs[0] = sgn;
        i++;
    }`

src/Longitude.cpp
```diff
@@ -102,15 +107,17 @@ const char *Longitude::formatStringForMeade(char *targetBuffer) const
...
     LOG(DEBUG_MEADE, "[LONGITUDE] Degs is %l, Mins is %l", degs, mins);
 
-    // Since internal storage is actual longitude, Meade is negated
+    char sgn;
     if (totalSeconds > 0)
     {
-        // Since we already inverted it when it was negative (by using ABS a few
-        // lines above here), we only invert it if it is positive.
-        degs = -degs;
+        sgn = '-';
```
peteasa, line 113: I think I got something wrong here.. when I checked the INDI logs I spotted 
Longitude 00:33:0.0 (0.55) (Longitude sign in carthography format) but this is not correct for my location .. the carthography longitude is -00:33:0 .. I am west of the prime meridian.  OAT does get the correct longitude it shows on the LCD display 0.54W 

I will do a bit more testing!

peteasa: well I did a quick test with all the debug on for INDI and this is what I get 
SCOPE	88.941888 sec	: <setSiteLongitude>
SCOPE	88.941930 sec	: CMD <:Sg000*32#>
SCOPE	88.964877 sec	: CMD <:Sg000*32#> successful.
....
INFO	88.986112 sec	: Site location in the mount updated to Latitude xxxxx Longitude  0:32:00.0 (0.533333) (Longitude sign in carthography format)
DEBUG	88.987411 sec	: Configuration successfully saved for GEOGRAPHIC_COORD.
INFO	88.987441 sec	: Observer location updated: Latitude xxxx Longitude  0:32:00.0 (0.53)

So my conclusion for this is that the INDI log statement might be incorrect since it is sending over the wire the correct west positive longitude.

I checked in skychart and it is correctly showing the longitude as -0:32:00 ie West 32 degrees and that is correct.

src/Longitude.cpp
peteasa: Yep, I also did a grep through the code to look for were formatString is used.  For Longitude formatStringForMeade is used rather than the basic formatString code.  I agree that it is rather dangerous to assume that in the future someone might not see this and pass something else in with *format, however there is no protection in formatStringImpl for sgn and format '{+}' being used at the same time so you could get '++' generated by formatStringImpl anyway.  So to fix that I would make a change in formatStringImpl...

PR #243, opened 2024-04-30
V1.13.3 Updates

ClutchplateDude:
- Added support for informational displays (no interactivity)
- Added implementation for a 128x64 OLED informational display
- Allowed ALT steps per revolution to be set directly
- Removed drift alignment and added ability to add it back with define
- Optimized string memory a little
- Lowered ESP32 second core priority
- Requires OATControl V1.1.2.0


ConfigurationValidation.hpp
```diff
@@ -34,6 +34,20 @@
...
 
+// Info Display validations
+#if defined(__AVR_ATmega2560__)
+    #if (INFO_DISPLAY_TYPE != INFO_DISPLAY_TYPE_NONE)
+        #if (DISPLAY_TYPE != DISPLAY_TYPE_NONE)
+            #error Unsupported configuration. Not enough memory to use LCD and OLED at the same time.
+        #endif
+        #if (FOCUS_STEPPER_TYPE != STEPPER_TYPE_NONE)
+            #if (BOARD == BOARD_AVR_MKS_GEN_L_V1) || (BOARD == BOARD_AVR_MKS_GEN_L_V2) || (BOARD == BOARD_AVR_MKS_GEN_L_V21)
+                #error Unsupported configuration. OLED requires I2C, which is on the E1 port on a MKS GenL, so a focuser cannot be used
```
julianneswinoga: The focuser doesn't need to be on E1, maybe we could check the pins? But we can only go from stepper to pin, not check if a pin is used anywhere... (afaik). :frowning_face: 

We should err on the side of not producing a false positive though, since it could be a valid configuration that we're blocking.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "esp8266-oled-ssd1306"]
+	path = esp8266-oled-ssd1306
+	url = https://github.com/ClutchplateDude/esp8266-oled-ssd1306
```
julianneswinoga: Instead of a submodule can we do a platformio git library? Like what we do for `avr-interrupt-stepper`. Submodules just tend to be a bit of a headache, so if we can avoid it that'd be nice

julianneswinoga (changes requested):

.gitmodules
ClutchplateDude: I have no idea how to do that, I just read in a PR a while back that we should do submodules :-D . Anyway, my git knowledge is only rudimentary. Can you point at a doc that explains how to do that?

ConfigurationValidation.hpp
ClutchplateDude: Ahh, yes, I see. You could put the focuser on another port if you're not using ALT AZ.... Should I change it to a warning? Should we generally just emit a warning to not use E1 when this is enabled?

.gitmodules
ClutchplateDude: Updated.

ConfigurationValidation.hpp
ClutchplateDude: Changed to warning.

julianneswinoga (approved):

PR #249, opened 2024-09-25
Update avr-interrupt-stepper library to version 0.0.3

andre-stefanov:
this version fixes the moveTo integer overflow which resulted in DEC runaway


ClutchplateDude (approved):
LGTM

PR #253, opened 2024-10-17
Fix git tag specification in pio lib dependencies

julianneswinoga:
`@` specifiers are not valid for git dependencies (only PlatformIO dependencies). Example 5 here: https://docs.platformio.org/en/latest/core/userguide/pkg/cmd_install.html#repository-git-hg-svn


ClutchplateDude (approved):
LGTM

PR #264, opened 2025-04-20
V1.13.16 Updates

ClutchplateDude:
- Limited refresh rate of InfoDisplay to:
   - 30fps when not slewing RA or DEC.
   - 5fps when slewing either RA or DEC axis.
   - off when slewing both RA and DEC axes.
- Ignore DEC Guide pulses when at DEC limits.

### DIFF
--- a/src/Mount.cpp
+++ b/src/Mount.cpp
@@ -357,7 +357,7 @@ void Mount::configureAZStepper(byte pin1, byte pin2, int maxSpeed, int maxAccele
     #ifdef NEW_STEPPER_LIB
     _stepperAZ = new StepperAzSlew(AccelStepper::DRIVER, pin1, pin2);
     #else
-    _stepperAZ    = new AccelStepper(AccelStepper::DRIVER, pin1, pin2);
+    _stepperAZ = new AccelStepper(AccelStepper::DRIVER, pin1, pin2);
     #endif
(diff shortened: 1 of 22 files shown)


src/Mount.cpp
```diff
@@ -3225,13 +3248,26 @@ void Mount::setupInfoDisplay()
...
+    int refreshRateHz = 30;
+    long now = millis();
+    if ((slewStatus() & (SLEWING_DEC | SLEWING_RA)) == (SLEWING_DEC | SLEWING_RA)) 
+    {
+        return;
+    }
+     else if (isSlewingRAorDEC()) 
+    {
+        refreshRateHz = 5;
+    }
```
julianneswinoga:
```suggestion
    int refreshRateHz = 30;
    const bool isSlewingRAandDEC = (slewStatus() & (SLEWING_DEC | SLEWING_RA)) == (SLEWING_DEC | SLEWING_RA);
    if (isSlewingRAandDEC) 
    {
        return;  // Do not update the display
    }
     else if (isSlewingRAorDEC()) 
    {
        refreshRateHz = 5;  // Update the display slower
    }
    const long now = millis();
```

julianneswinoga (dismissed):
Minor change request

julianneswinoga (approved):

PR #274, opened 2026-01-13
V1.13.18 - Updates

ClutchplateDude:
- Merged oae-fw branch so that all mounts are in main branch (develop) now.
- Added more output to the InfoDisplay during boot and fixed a bug in console mode.
- Fixed a bug that was causing the firmware to hang after a slew, if a :Q# command was issued (thanks to user c3n for tracking it down, pun intended).


src/b_setup.hpp
```diff
@@ -118,6 +134,8 @@ void setup()
 #if TEST_VERIFY_MODE == 1
     #ifdef OAM
     Serial.print(F("Booting OAM Firmware "));
+    #elif OAE
```
andre-stefanov: the first condition is `#ifdef OAM` which equals `#if defined(OAM)`. the second one is "just" `#elif OAE`. This could cause issues if `OAE` flag is defined without a value.
```suggestion
    #elif defined(OAE)
```

src/b_setup.hpp
```diff
@@ -134,32 +152,46 @@ void setup()
...
     mount.setupInfoDisplay();
+    addConsoleText(F("BOOTING " VERSION));
     LOG(DEBUG_ANY, "[SYSTEM]: OLED info screen ready!");
-    mount.getInfoDisplay()->addConsoleText(F("BOOTING " VERSION), false);
 #endif
 
 #if USE_GPS == 1
+    LOG(DEBUG_ANY, "[SYSTEM]: Initializing GPS...");
+    #if (INFO_DISPLAY_TYPE != INFO_DISPLAY_TYPE_NONE)
+    int gpsLine = addConsoleText(F("Initialize GPS..."), false);
```
andre-stefanov: what is the argument `false`? `addConsoleText(String text)` in b_setup has just one parameter. `InfoDisplayRender#addConsoleText` on the other side has two parameters ... but second one does not seem to be used anywhere.

src/SSD1306_128x64_Display.hpp
andre-stefanov: i know it is not part of the PR, but we should check for overflow here. `_curLine++` has to be smaller than `MAX_CONSOLE_LINES`

Otherwise we need a strategy ... maybe one of these:
- FIFO like: adding a line would drop the first one. This is not that sexy to implement since we would need to shift all the array entries down before adding new line. It would be probably easier by using a ring buffer as structure.
- drop new line and return an error

ConfigurationValidation.hpp
```diff
@@ -89,6 +89,8 @@
             #error AZ driver address for DRIVER_TYPE_TMC2209_UART not specified.
         #endif
     #endif
+#elif defined(BOARD_OAE_V1)
```
andre-stefanov: This check will not work as expected. `BOARD_OAE_V1` is always defined in [Constants.hpp](https://github.com/OpenAstroTech/OpenAstroTracker-Firmware/pull/274/changes#diff-83a6144ef3c6a82e9a2ebb847f84ff83a19a26fc04eeec521281b85272e3afd6R21). Because of this the `#else` block will be never reached.

```suggestion
#elif (BOARD == BOARD_OAE_V1)
```

ConfigurationValidation.hpp
```diff
@@ -110,6 +112,9 @@
         #endif
     #endif
 
+#elif defined(BOARD_OAE_V1)
```
andre-stefanov: This check will not work as expected. `BOARD_OAE_V1` is always defined in [Constants.hpp](https://github.com/OpenAstroTech/OpenAstroTracker-Firmware/pull/274/changes#diff-83a6144ef3c6a82e9a2ebb847f84ff83a19a26fc04eeec521281b85272e3afd6R21). Because of this the `#else` block will be never reached.

```suggestion
#elif (BOARD == BOARD_OAE_V1)
```

src/SSD1306_128x64_Display.hpp
```diff
@@ -92,20 +95,29 @@ class SDD1306OLED128x64 : public InfoDisplayRender
 
             // Name on the right
             display->setFont(Bitmap5x7);
-#ifdef OAM
+#ifdef OAE
+            display->drawString(32, 6, F("OpenAstroExplorer"));
+#elif defined(OAM)
             display->drawString(32, 6, F("OpenAstroMount"));
+#elif defined(OAE)
+            display->drawString(32, 6, F("OpenAstroExplorer"));
```
andre-stefanov: Duplicate of lines 98-99

boards/ESP32_ESP32DEV/pins_OAE_V1.hpp
```diff
@@ -0,0 +1,134 @@
...
+ * @brief a pins configuration file for an OAE board v1.0
+ */
+
+#pragma once
+
+/**
+ * @brief a pins configuration file for an ESP32-based OAT.
+ */
+
+#pragma once
```
andre-stefanov: duplicate of lines 1-5

boards/ESP32_ESP32DEV/pins_OAE_V1.hpp
```diff
@@ -0,0 +1,134 @@
...
+#endif
+#ifndef RA_STEPPER_SPR
+    #define RA_STEPPER_SPR (400 * 9)  // change to (200 * 9) for 1.8° stepper
+#endif
+#ifndef DEC_STEPPER_SPR
+    #define DEC_STEPPER_SPR (200 * 50 * 4.5f)  // change to (200 * 9) for 1.8° stepper
+#endif
+#ifndef DEC_PULLEY_TEETH
+    #define DEC_PULLEY_TEETH 1
+#endif
```
andre-stefanov: I recommend adding an empty new line at the end of this file. This makes git diffs less noisy if we add more content to the end of the file later.

andre-stefanov (changes requested):

src/b_setup.hpp
ClutchplateDude: Good catch!

src/b_setup.hpp
ClutchplateDude: Damn, you are on a roll! Yeah, that was a leftover from when I replaced all the calls...

src/SSD1306_128x64_Display.hpp
ClutchplateDude: This is way out of date file.... urgh... let me check what's going on here.

src/SSD1306_128x64_Display.hpp
ClutchplateDude: Something is wrong in this PR, these are old files...

src/InfoDisplayRender.hpp
```diff
@@ -1,17 +1,116 @@
 #pragma once
 #include <Arduino.h>
-
-class Mount;
+#include <String.h>
```
andre-stefanov: `String.h` cant be found during builds

src/b_setup.hpp
```diff
@@ -174,12 +206,25 @@ void setup()
...
+    LOG(DEBUG_ANY, "[SYSTEM]: Initializing TMC2209 UART pins and Serial port for RA...");
     // include TMC2209 UART pins
+    #if defined(RA_DIAG_PIN)
     pinMode(RA_DIAG_PIN, INPUT);
+    #endif
+
     #ifdef RA_SERIAL_PORT
-    RA_SERIAL_PORT.begin(57600);  // Start HardwareSerial comms with driver
+        #ifdef OAE
+    RA_SERIAL_PORT.begin(57600, SERIAL_8N1, RA_TX_PIN, RA_RX_PIN);
```
andre-stefanov: are the RX/TX pins swapped here? 

```c++
void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, int8_t txPin, bool invert, unsigned long timeout_ms, uint8_t rxfifo_full_thrhd)
```

src/b_setup.hpp
```diff
@@ -193,15 +238,27 @@ void setup()
...
+
 #if DEC_DRIVER_TYPE == DRIVER_TYPE_TMC2209_UART
+    LOG(DEBUG_ANY, "[SYSTEM]: Initializing TMC2209 UART pins and Serial port for DEC...");
     // include TMC2209 UART pins
+    #if defined(DEC_DIAG_PIN)
     pinMode(DEC_DIAG_PIN, INPUT);
+    #endif
     #ifdef DEC_SERIAL_PORT
+        #ifdef OAE
+    DEC_SERIAL_PORT.begin(57600, SERIAL_8N1, DEC_TX_PIN, DEC_RX_PIN);
```
andre-stefanov: are the RX/TX pins swapped here? 

```c++
void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, int8_t txPin, bool invert, unsigned long timeout_ms, uint8_t rxfifo_full_thrhd)
```

src/Mount.cpp
```diff
@@ -569,6 +581,10 @@ void Mount::configureDECdriver(Stream *serial, float rsense, byte driveraddress,
         #endif
     LOG(DEBUG_STEPPERS, "[MOUNT]: Requested DEC motor rms_current: %d mA", rmscurrent);
     _driverDEC->rms_current(rmscurrent, 1.0f);  //holdMultiplier = 1 to set ihold = irun
+    _driverDEC->pdn_disable(1);
+    _driverDEC->ihold(31);
+    _driverDEC->iholddelay(15);
+    _driverDEC->TPOWERDOWN(255);
```
andre-stefanov, line 587: are these changes specific to OAE or is this a general change? it will influence other mounts/boards as well here. If i get it correctly, we set the hold current here to maximum ... right?

andre-stefanov (changes requested):

src/InfoDisplayRender.hpp
ClutchplateDude: This was removed

src/b_setup.hpp
ClutchplateDude: Oh, wow. This is Fabians code :-) But this would mean RA UART comms would never work, right? But they do?

andre-stefanov: well ... i have no idea :D i worked out of the box for me. so either defaults (without UART) are luckily also suitable OR maybe he named the pins the wrong way around?

andre-stefanov (approved):

src/Mount.cpp
ClutchplateDude: General change to improve DEC guide performance

PR #277, opened 2026-01-30
Add conditional definition for SW_SERIAL_UART

jrhuerta:
Define SW_SERIAL_UART if not already defined.
Avoid warnings when configuring ramps board to use Serial3 for TMC2209 drivers in UART mode in Configuration_local.hpp
```c++
#define SW_SERIAL_UART 0
#define RA_SERIAL_PORT Serial3
#define RA_DRIVER_ADDRESS 0b00
#define DEC_SERIAL_PORT Serial3
#define DEC_DRIVER_ADDRESS 0b01
```

### DIFF
--- a/boards/RAMPS/pins_RAMPS.hpp
+++ b/boards/RAMPS/pins_RAMPS.hpp
@@ -51,7 +51,9 @@
     #define DEC_DRIVER_ADDRESS 0b00
 #endif
 
-#define SW_SERIAL_UART 1
+#ifndef SW_SERIAL_UART
+    #define SW_SERIAL_UART 1
+#endif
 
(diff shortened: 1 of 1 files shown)


ClutchplateDude (approved):
LGTM

jrhuerta:
Thanks for taking a look at this PR.

I noticed the changelog check is failing. I’ve seen cases where updates are bundled and the version is bumped afterward, please let me know if there’s anything I should address on my side, or if I should just wait.

ClutchplateDude:
Yeah, generally, PRs require a version update and a changelog update.... But with such trivial changes, I think we can circumvent that rule...

PR #278, opened 2026-02-07
Enhance GPS and RA/DEC serial port initialization for ESP32 compatibility

jrhuerta:
When using an ESP32 based board, allow redefinition of the serial port pins in the configuration.


ClutchplateDude (approved):
LGTM

PR #293, opened 2026-05-27
refactor: moved more logic components into core

andre-stefanov:
### Summary

Extracts the pure (non-Arduino) arithmetic subsets of the firmware's core data types and algorithms into core, making them host-testable via the `native` PlatformIO environment. The original src files become thin overlays that inherit from the core versions and add only Arduino-dependent methods (`ParseFromMeade`, `ToString`). Zero behavior change — all 5 boards build green.

### Architecture

```
src/core/
├── types/                     # Pure data containers (no Arduino)
│   ├── DayTime.hpp/.cpp       # 24h time arithmetic, formatString
│   ├── RightAscension.hpp/.cpp# 0–24h RA coordinate (wrapping)
│   ├── Declination.hpp/.cpp   # ±180° DEC coordinate (clamping)
│   ├── Latitude.hpp/.cpp      # ±90° latitude (clamping)
│   └── Longitude.hpp/.cpp     # ±180° longitude (wrapping)
├── SiderealClock.hpp/.cpp     # LST/HA calculation (theta, delta-JD)
├── CalendarMath.hpp/.cpp      # Date arithmetic with leap-year rules
├── CoordinateMath.hpp/.cpp    # Meridian-flip solution selector
├── CoordinateFormatter.hpp/.cpp# RA/DEC char-buffer formatting
└── EepromLayout.hpp           # Magic markers, flags, layout constants
```

Each src originals (`DayTime`, `Declination`, `Latitude`, `Longitude`) is now a thin overlay:

```cpp
// src/DayTime.hpp
class DayTime : public core::DayTime {
    using core::DayTime::DayTime;  // inherit constructors
    virtual const char *ToString() const;
    static DayTime ParseFromMeade(String const &s);
};
```

### New files

| # | File |
|---|------|
| 6 | `src/core/types/{DayTime,RightAscension,Declination,Latitude,Longitude}.{hpp,cpp}` |
| 4 | `src/core/{SiderealClock,CalendarMath,CoordinateMath,CoordinateFormatter}.{hpp,cpp}` |
| 1 | EepromLayout.hpp |
| 2 | `src/{RightAscension}.{hpp,cpp}` (new overlay) |
| 10 | `unit_tests/test_core/{test_sidereal,test_calendar,test_coordinate_math,test_coordinate_formatter,test_eeprom_layout,types/test_daytime,types/test_declination,types/test_latitude,types/test_longitude,types/test_right_ascension}.cpp` |

### Modified files

| # | File | Change |
|---|------|--------|
| 8 | `src/{DayTime,Declination,Latitude,Longitude}.{hpp,cpp}` | Converted to thin overlays; pure methods removed |
| 1 | Sidereal.cpp | Delegates to `core::SiderealClock` |

### Test results

```
258 test cases: 258 succeeded (188 existing + 70 new)
```
All 5 board matrix builds green (`ramps`, `mksgenlv21`, `mksgenlv2`, `mksgenlv1`, `esp32`, `oaeboardv1`).

### Key decisions

- **Extract-before-test:** src root files transitively include `<Arduino.h>` via `inc/Globals.hpp` and are untestable on native. Pure logic is extracted into `core/` first, then tested.
- **Thin overlays preserve back-compat:** Originals `#include` the core version and add only `String`-dependent methods. No flag day.
- **`core` namespace:** All extracted types live in `namespace core` to avoid collisions with the overlay classes in the global namespace.
- **`core::RightAscension`** was added as a first-class RA coordinate type alongside `Declination`, replacing `DayTime` in `CoordinateFormatter::formatRA()`. Inheritance from `DayTime` is acknowledged as a transitional convenience — a composition-based design is planned for Phase 2.

### What's next (Phase 2)

Introduce HAL interfaces (`IStepperMotor`, `ISystemClock`, `IEeprom`, …), domain ports (`IClock`, `IPersistentStore`, `IStepperAxis`, …), and adapters that bind ports to HAL. Wire `Mount` to use injected port pointers instead of concrete types.


src/hal/EepromLayout.hpp
andre-stefanov, line 1: now i realize. EepromLayout should not be part of the core. it describes how data is persisted and not what the domain does.

andre-stefanov: A better place would be under `/hal` alongside future `IEeprom`

andre-stefanov: moved

unit_tests/test_core/types/test_daytime.cpp
```diff
@@ -0,0 +1,180 @@
...
+{
+    DayTime dt(5, 15, 30);
+    int h, m, s;
+    dt.getTime(h, m, s);
+    EXPECT_EQ(5, h);
+    EXPECT_EQ(15, m);
+    EXPECT_EQ(30, s);
+}
+
+TEST(DayTimeTest, GetTimeRefNegative)
```
ClutchplateDude, line 170: Need a test for -2, 31, 18.

unit_tests/test_core/types/test_daytime.cpp
```diff
@@ -0,0 +1,180 @@
...
+    DayTime dt(2, 30, 0);
+    EXPECT_FLOAT_EQ(2.5f, dt.getTotalHours());
+}
+
+TEST(DayTimeTest, GetTotalMinutes)
+{
+    DayTime dt(1, 30, 0);
+    EXPECT_FLOAT_EQ(90.0f, dt.getTotalMinutes());
+}
+
```
ClutchplateDude, line 49: Need test for -2, 45, 0 and check totalMinutes to be -165.

ClutchplateDude (changes requested):

unit_tests/test_core/types/test_daytime.cpp
andre-stefanov: done

unit_tests/test_core/types/test_daytime.cpp
andre-stefanov: done

ClutchplateDude (approved):

PR #295, opened 2026-07-20
Fixes and improvements for ESP32 in AP mode:

tobiashennerbichler:
- Now correctly reports the device's IP in AP mode for UDP discovery requests. Previously reported as 0.0.0.0.
- Split logic between AP mode and Infra modes since WiFi.status() returns a meaningless value in AP mode. AP mode now immediately establishes TCP/UDP servers. Server establishment moved to own function.
- Also adds null-pointer guards to tcp/udp loop while servers are not established yet. Not relevant for AP mode, but included here due to simplicity.

Tested on ESP32 in WIFI_MODE_AP_ONLY. This PR does NOT address the infrastructure modes or the WIFI_MODE_ATTEMPT_INFRASTRUCTURE_FAIL_TO_AP mode in particular, which has some issues in the failover trigger. Local changes are ready and can be included in future PRs if desired.

### DIFF
--- a/src/WifiControl.cpp
+++ b/src/WifiControl.cpp
@@ -109,13 +109,21 @@ String WifiControl::getStatus()
         result += "Infra-Fail-To-AP,";
     }
 
-    result += wifiStatus(WiFi.status()) + ",";
+    if(WIFI_MODE != WIFI_MODE_AP_ONLY)
+    {
+        result += wifiStatus(WiFi.status()) + ",";
+    }
     #if defined(ESP32)
     result += WiFi.getHostname();
+    result += ",";
     #endif
 
-    result += "," + getIP() + ":" + WIFI_PORT;
-    result += "," + String(WIFI_INFRASTRUCTURE_MODE_SSID) + "," + String(WIFI_HOSTNAME);
+    result += getIP() + ":" + WIFI_PORT + ",";
+    if (WIFI_MODE != WIFI_MODE_AP_ONLY)
+    {
+        result += String(WIFI_INFRASTRUCTURE_MODE_SSID) + ",";
+    }
+    result += String(WIFI_HOSTNAME);
 
     return result;
 }
@@ -129,7 +137,7 @@ void WifiControl::establishServers()
 
     delete _udp;
     _udp = new WiFiUDP();
-    _udp->begin(4031);
+    _udp->begin(WIFI_UDP_DISCOVERY_PORT);
 }
 
 String WifiControl::getIP()
@@ -161,12 +169,11 @@ void WifiControl::loop()
                     WiFi.localIP().toString().c_str(),
                     WIFI_PORT);
             }
-
-            if (_status != WL_CONNECTED)
-            {
-                infraToAPFailover();
-                return;
-            }
+        }
+        if (_status != WL_CONNECTED)
+        {
+            infraToAPFailover();
+            return;
         }
     }
 
@@ -263,7 +270,7 @@ void WifiControl::udpLoop()
         incomingPacket[lookingFor.length()] = 0;
         if (lookingFor.equalsIgnoreCase(incomingPacket))
         {
-            _udp->beginPacket(_udp->remoteIP(), 4031);
+            _udp->beginPacket(_udp->remoteIP(), WIFI_UDP_DISCOVERY_PORT);
             /*unsigned char bytes[255];
             reply.getBytes(bytes, 255);
             _udp->write(bytes, reply.length());*/
--- a/src/WifiControl.cpp
+++ b/src/WifiControl.cpp
@@ -61,8 +61,11 @@ void WifiControl::startAccessPointMode()
     WiFi.setHostname(WIFI_HOSTNAME);
     #endif
 
-    WiFi.softAP(WIFI_HOSTNAME, WIFI_AP_MODE_WPAKEY);
+    WiFi.mode(WIFI_AP);
     WiFi.softAPConfig(local_ip, gateway, subnet);
+    WiFi.softAP(WIFI_HOSTNAME, WIFI_AP_MODE_WPAKEY);
+
+    establishServers();
 }
 
 String wifiStatus(int status)
@@ -111,10 +114,29 @@ String WifiControl::getStatus()
     result += WiFi.getHostname();
     #endif
 
-    result += "," + WiFi.localIP().toString() + ":" + WIFI_PORT;
+    result += "," + getIP() + ":" + WIFI_PORT;
     result += "," + String(WIFI_INFRASTRUCTURE_MODE_SSID) + "," + String(WIFI_HOSTNAME);
 
     return result;
+}
+
+void WifiControl::establishServers()
+{
+    delete _tcpServer;
+    _tcpServer = new WiFiServer(WIFI_PORT);
+    _tcpServer->begin();
+    _tcpServer->setNoDelay(true);
+
+    delete _udp;
+    _udp = new WiFiUDP();
+    _udp->begin(4031);
+}
+
+String WifiControl::getIP()
+{
+    return WIFI_MODE == WIFI_MODE_DISABLED  ? "NONE"
+           : WIFI_MODE == WIFI_MODE_AP_ONLY ? WiFi.softAPIP().toString()
+                                            : WiFi.localIP().toString();
 }
 
 void WifiControl::loop()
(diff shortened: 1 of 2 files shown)


ClutchplateDude:
# Code Review — "Fixes and improvements for ESP32 in AP mode"

**Scope:** ESP32 Wi-Fi in `WIFI_MODE_AP_ONLY` (infra / `WIFI_MODE_ATTEMPT_INFRASTRUCTURE_FAIL_TO_AP` explicitly excluded per the PR description). `src/WifiControl.cpp` / `.hpp`, +86 / −57.

## Overview

- **`startAccessPointMode()`** — adds `WiFi.mode(WIFI_AP)` and reorders to `mode → softAPConfig → softAP`, then calls the new `establishServers()`.
- **`establishServers()`** — extracted helper that (re)creates the TCP/UDP servers (was inline in `loop()`).
- **`getIP()`** — returns `softAPIP()` in AP-only, `localIP()` otherwise, `"NONE"` when disabled; used by `getStatus()` and `udpLoop()`.
- **`loop()`** — the status-tracking / server-setup / failover block is now gated by `if (WIFI_MODE != WIFI_MODE_AP_ONLY)`, so AP mode skips it entirely.
- **`tcpLoop()` / `udpLoop()`** — null-pointer guards (`if (!_tcpServer) return;` / `if (!_udp) return;`); `available()` → `accept()`.

The core AP-mode fix is correct and complete, and the refactor is clean. One real regression for the failover path, plus a couple of minor loose ends.

## Strengths

- **Root-cause fix is right.** Adding `WiFi.mode(WIFI_AP)` before `softAP()`/`softAPConfig()` is the documented ESP32 order; without `WIFI_AP` set, `softAPIP()` returns `0.0.0.0` — exactly the symptom described.
- **Null guards are the load-bearing piece of the refactor.** Because `tcpLoop`/`udpLoop` now no-op until servers exist, it becomes safe to stop early-returning out of `loop()` when not yet connected. Internally coherent design.
- **`establishServers()` extraction** removes duplication (the same code ran in `loop()` and now also runs in `startAccessPointMode()`, which is itself called from `infraToAPFailover()` — so failover correctly re-establishes servers too).
- **`_status = WL_DISCONNECTED` member init** fixes a previously-indeterminate `wl_status_t`.
- **`available()` → `accept()`** is a correct modernization: it accepts on connect rather than waiting for data, then `while (client.available())` waits for bytes. Slightly more correct than before.
- **Style** matches house conventions (Allman braces, 4-space indent, right-aligned pointers, `LOG(DEBUG_WIFI, …)`).

## Issues & Risks

### 1. [Important] Failover is now edge-triggered — likely regression for `WIFI_MODE_ATTEMPT_INFRASTRUCTURE_FAIL_TO_AP`

This is the headline finding. The old `loop()` polled `infraToAPFailover()` **every iteration** whenever `_status != WL_CONNECTED`:

```cpp
// OLD
_mount->loop();
if (_status != WL_CONNECTED) {
    infraToAPFailover();   // called every loop — timeout gets re-evaluated each tick
    return;
}
```

The new code only calls it **inside the status-change block**:

```cpp
// NEW
if (WIFI_MODE != WIFI_MODE_AP_ONLY) {
    if (_status != WiFi.status()) {
        _status = WiFi.status();
        ...
        if (_status != WL_CONNECTED) { infraToAPFailover(); return; }   // only on a transition
    }
}
```

`infraToAPFailover()` is timeout-gated (`_infraStart + _infraWait < millis()`). After a failed infra attempt, `WiFi.status()` typically settles at `WL_DISCONNECTED` and stops changing. At that point `_status == WiFi.status()` forever, the change-block never re-enters, and **the timeout is never re-evaluated** — so the fall-over-to-AP never fires. The device hangs on infra indefinitely.

The old every-loop poll is what made the timeout actually expire. The author notes FAIL_TO_AP "has some issues" and is out of scope, but this PR doesn't just leave it alone — it removes the mechanism that let the existing timeout work.

**Suggested fix** — keep a periodic (not edge) failover poll for the non-AP modes, e.g.:

```cpp
if (WIFI_MODE != WIFI_MODE_AP_ONLY) {
    if (_status != WiFi.status()) {
        _status = WiFi.status();
        /* log */
        if (_status == WL_CONNECTED) establishServers();
    }
    if (_status != WL_CONNECTED) { infraToAPFailover(); return; }   // back to every-loop
}
```

The null guards mean re-adding the every-loop `return` is still safe.

### 2. [Minor] `getStatus()` still reports a meaningless Wi-Fi status in AP mode

`getStatus()` (unchanged) still calls `wifiStatus(WiFi.status())`. By the PR's own rationale (`WiFi.status()` is meaningless in AP mode), the status field in AP-only will show an arbitrary string. The PR fixed the IP half of this but not the status half. Consider returning a fixed connected/"Access Point" status for AP-only.

### 3. [Minor] `getIP()` returns the wrong IP after a FAIL_TO_AP failover

`getIP()` keys off the compile-time `WIFI_MODE`, so for `WIFI_MODE_ATTEMPT_INFRASTRUCTURE_FAIL_TO_AP` it always returns `WiFi.localIP()` — even after `infraToAPFailover()` has switched the radio to AP. The UDP discovery reply and `getStatus()` would then advertise `0.0.0.0`/the stale STA IP in the failed-over state — the exact class of bug this PR is fixing for AP-only. Acceptable given the stated scope, but worth a `// TODO` or tracking runtime mode.

### 4. [Nit] Magic number `4031` now appears in two places

`4031` (UDP discovery port) was already a bare literal; `establishServers()` duplicates it alongside the original `udpLoop()` use. A named constant (e.g. `WIFI_DISCOVERY_PORT`) would prevent the two from drifting. Pre-existing debt — optional.

## Other Dimensions

- **Correctness (non-infra):** AP-only path is sound end-to-end. `establishServers()` `delete`s before `new`-ing (null-delete is safe), so re-establishment on failover doesn't leak.
- **Performance:** Negligible — one `WiFi.mode()` at startup; one extra branch per loop for the null guards.
- **Security:** No new surface. `readStringUntil('#')` into a heap `String` is pre-existing and unchanged.
- **Test coverage:** No unit tests added, but `WifiControl` is hardware-bound and the native suite doesn't cover Wi-Fi; manual ESP32 testing in AP-only is noted. Acceptable.
- **Conventions:** Compliant with the project style rules; should pass clang-format and `-Werror`.

## Verdict

**Approve with one requested change:** the AP-only fix is correct and well-executed, but issue #1 (failover polling) is a concrete behavioral regression for `FAIL_TO_AP` that's worth fixing or explicitly documenting as known-broken in this PR. Issues #2–#4 are minor/cosmetic.

src/WifiControl.cpp
```diff
@@ -172,57 +190,66 @@ void WifiControl::infraToAPFailover()
...
         {
-            LOG(DEBUG_WIFI, "[WIFITCP]: Available bytes %d. Peeking.", client.available());
+            client.read();
+            LOG(DEBUG_WIFI, "[WIFITCP]: Query <-- Handshake request");
+            client.write("P");
+            LOG(DEBUG_WIFI, "[WIFITCP]: Reply --> P (polar mode)");
+        }
+        else
+        {
+            String cmd = client.readStringUntil('#');
```
ClutchplateDude: **Blocking read in `tcpLoop()` - please mirror the non-blocking serial pattern**

`client.readStringUntil('#')` blocks until `#` arrives or the `Stream` timeout (1 s default) elapses. Since `tcpLoop()` is entered as soon as *any* bytes are available, a client sending a partial command (no trailing `#`) stalls `WifiControl::loop()` - and the `_mount->loop()` it's meant to service - for up to a second. On AVR that freezes the main loop; on ESP32 the Core-1 loop stalls (steppers keep ticking on Core 0, but guiding/status updates don't).

The serial path already does this right - see `processSerialData()` in `src/f_serial.hpp`: it drains `while (Serial.available() > 0)` one byte at a time into a fixed `char buffer[]`, dispatches on `#`, guards against overflow, and calls `mount.loop()` per byte. `tcpLoop()` should follow the same accumulate-until-`#` approach.

I know this is pre-existing and not something you added here, but since `tcpLoop()` is being rewritten in this PR it'd be a good moment to align it.

ClutchplateDude (approved):
Thanks for your submission and fixes!

rbhbokka:
I ran into the same dead port in AP-only mode and had arrived at nearly the same fix, so I'm dropping mine in favour of this one — `establishServers()` and the `getIP()` split are the right shape.

Two notes on the failover mode you already flagged as out of scope, in case they save you time later.

### 1. `WIFI_MODE_ATTEMPT_INFRASTRUCTURE_FAIL_TO_AP` still can't serve a client once it has failed over

In `loop()` at `e456530`:

```cpp
if (WIFI_MODE != WIFI_MODE_AP_ONLY)
{
    ...
    if (_status != WL_CONNECTED)
    {
        infraToAPFailover();
        return;              // <--
    }
}

_mount->loop();
tcpLoop();
udpLoop();
```

`WIFI_MODE` is a compile-time constant, so in failover mode the outer branch is taken forever. Once `infraToAPFailover()` has called `startAccessPointMode()`, the station is disconnected and `WiFi.status()` never reports `WL_CONNECTED` again — so `_status != WL_CONNECTED` stays true and the `return` fires on every subsequent pass. `_infraStart` is zeroed, so the failover itself correctly becomes a no-op, but `tcpLoop()` / `udpLoop()` are never reached again. The soft-AP is up and `establishServers()` has bound the port, yet no client is ever serviced.

To be clear, this is not introduced here — the same early return is on `develop` today. It just means the mode stays unusable after this change, which matches your note in the description.

What makes it awkward is that one condition is answering two different questions: *"should we attempt failover?"* (policy) and *"are the servers up?"* (lifecycle). Separating them is a small edit:

```diff
     if (WIFI_MODE != WIFI_MODE_AP_ONLY)
     {
         if (_status != WiFi.status())
         {
             ...
         }
         if (_status != WL_CONNECTED)
         {
             infraToAPFailover();
-            return;
         }
     }
 
+    if (_tcpServer == nullptr || _udp == nullptr)
+    {
+        return;
+    }
+
     _mount->loop();
     tcpLoop();
     udpLoop();
```

`infraToAPFailover()` is already self-guarding on `_infraStart != 0`, so it stays a no-op in AP-only mode and after a completed failover, and the null check becomes a pure lifecycle guard. Your `if (!_tcpServer) return;` / `if (!_udp) return;` guards inside the two loops already cover the dereference, so this is about reachability rather than safety.

### 2. Minor

`_mount->loop()` has moved below the early return, where it previously sat above it. `serialLoop()` and `loopButtons()` also call it so the mount still ticks, but it is a behaviour change that isn't in the description.

Happy to open a follow-up implementing the above on top of this once it lands, rather than complicating this PR.

PR #291, opened 2026-05-19
refactor: Enhance Meade command parsing and response handling

andre-stefanov:
Refactor the Meade command parser and response system for improved organization and maintainability. Introduce new command handlers, enhance unit test coverage, and update documentation to clarify layer responsibilities. This update aims to streamline command processing and ensure robust handling of various Meade commands.

PR #296, opened 2026-08-05
Fix DEC and longitude conversions in Meade commands

GJayLG:
After #291 the Meade DEC handlers stopped converting between celestial
declination and the internal "degrees from pole" representation. A goto
to +26 36'08" was treated as 26.6 from the pole instead of 63.4, so
the mount pointed at the wrong part of the sky, and every platesolve
sync made the error worse.

The same refactor dropped the sign handling for longitude on the Meade wire east is negative, but the mount stores cartographic longitude
(east positive). That gave a wrong LST for observers east of Greenwich

### DIFF
--- a/src/MeadeCommandProcessor.cpp
+++ b/src/MeadeCommandProcessor.cpp
@@ -97,12 +97,27 @@ meade::RaCoordinate raFrom(const DayTime &t)
     };
 }
 
+Declination decToInternal(const meade::DecCoordinate &d)
+{
+    long seconds = labs(static_cast<long>(d.degrees)) * 3600L + d.minutes * 60L + d.seconds;
+    return Declination::FromSeconds(d.degrees < 0 ? -seconds : seconds);
+}
+
 meade::DecCoordinate decFrom(const Declination &d)
 {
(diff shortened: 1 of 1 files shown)


rbhbokka:
Heads-up about a merge hazard rather than a review comment, since #304 touches the same thing.

This and #304 shouldn't both land. #296 negates the longitude in `MeadeCommandProcessor::onSetSiteLongitude`, #304 does it in the parser. Different files, so git merges them happily and the result negates twice: the reader gives -121°53', then the handler flips it back to +121°53' East. No conflict, and no test that would catch it.

Three other things for whoever picks a direction:

- This reintroduces the sign-of-zero problem on both sides (`getTotalHours() > 0`, and `Longitude(-degrees, ...)` where `sign(0)` is 1), which is what #305 is removing.
- It doesn't touch `readLongitude`, so the unsigned `:Sg121*53#` that INDI actually sends still won't parse.
- The DEC half predates #300 and #302, and it conflicts with #305's struct change.

The getter-negation idea here is the right one and is what #304 does now, so credit where it's due if this ends up superseded.

